PR #3, opened 2023-04-27
EXPLORE EXTRACTED DATA: As an Open Source user, I want to be able to browse the data that I have extracted, so that I can make decisions about how to transform it. 

richardmatthewsdev:
**Acceptance Criteria**

- User can investigate extracted data
- User can navigate large data sets easily & safely
- Key information/ metrics for extracted data are presented
- Extraction error is displayed

**Notes**
- Design required.

**Key information/ metrics**
- Date of Extraction (+ duration?)
- Number of records
- Filesize?
- Transformations associated?
- ?


app/views/jobs/show.html.erb
```diff
@@ -21,86 +21,102 @@
...
+<ul class="list-group list-group-horizontal my-4">
+  <li class="list-group-item">Status: <%= job_status_text(@job) %></li>
+
+  <%- if @job.duration_seconds.present? %>
+    <li class="list-group-item">Duration: <%= ActiveSupport::Duration.build(@job.duration_seconds).inspect %></li>
+  <% end %>
+    <li class="list-group-item">Filesize: <%= number_to_human_size(@job.extraction_folder_size_in_bytes) %></li>
+    <li class="list-group-item">Records Per Page: <%= @extraction_definition.per_page %></li>
+    <li class="list-group-item">Total Pages: <%= @job.documents.total_pages %></li>
+    <li class="list-group-item">Total Records: <%= @extraction_definition.per_page * @job.documents.total_pages %></li>
```
paul-mesnilgrente, line 33: That's a bit misleading here, the count won't right if all pages are not filled.

richardmatthewsdev: Fair call, I'll update it to say that it's an estimate and add a tooltip for how it's generated

lun-boost (approved):
Looks pretty awesome!!! Thanks for walking me through it 🤠🙏

app/models/job.rb
```diff
@@ -12,39 +15,81 @@ class Job < ApplicationRecord
 
   validates :status, presence: true, inclusion: { in: STATUSES }
   validates :kind, presence: true, inclusion: { in: KINDS }
+  validates :end_time, comparison: { greater_than: :start_time }, if: ->{ end_time.present? }
 
   STATUSES.each do |status_name|
+    # A helper method to determine the status of a job
```
motizuki, line 21: is this the same thing as defining the status as enum?

app/models/job.rb
```diff
@@ -12,39 +15,81 @@ class Job < ApplicationRecord
...
+    return if start_time.blank? || end_time.blank?
+
+    end_time - start_time
+  end
+
+  # Returns the size of the extraction folder in bytes 
+  #
+  # @return Integer
+  def extraction_folder_size_in_bytes
+    `du #{extraction_folder} | cut -f1`.to_i
```
motizuki, line 93: i guess we need to be careful to not execute unexpected thigs in the extraction folder variable `du /path | rm -rf /`

motizuki: not sure if this is an input that comes from the frontend though, it might be ok?

motizuki (approved):

app/models/job.rb
richardmatthewsdev: Yeah it is! I'll switch it over to be an enum

app/models/job.rb
richardmatthewsdev: Yeah they can't define the extraction folder, it's worked out by the system :)

PR #4, opened 2023-05-01
CREATE TRANSFORMATION : As an Open Source User, I want to be able to set up a new transformation using my extracted data, so that I can make use of it in the way I need.

richardmatthewsdev:
**Acceptance Criteria**
- User can create a new transformation
- User can select extracted data
- User can specify record selector
- User can preview output of record selection


isabel-anastasiadis-boost (commented):
Would it be good to install `codeclimate_diff` so that it can be run? I was curious if there were any regressions.

I ran the audit on it, and it returned:


NAME                 | ORGANISATION | REPO_HOST  | RELATIVE_PATH | FOLDER | MAIN_BRANCH           | SIMILARITY_SCORE_PER_FILE | ABC_METHOD_AVERAGE | CODE_SMELLS_PER_FILE | TOTAL_SCORE
---------------------|--------------|------------|---------------|--------|-----------------------|---------------------------|--------------------|----------------------|------------
supplejack_harvester | DigitalNZ    | github.com | ../           | app    | create-transformation | 2.74                      | 6.1                | 0.41                 | 9.25

main currently looks like this: 

NAME                 | ORGANISATION | REPO_HOST  | RELATIVE_PATH | FOLDER | SIMILARITY_SCORE_PER_FILE | ABC_METHOD_AVERAGE | CODE_SMELLS_PER_FILE | TOTAL_SCORE
---------------------|--------------|------------|---------------|--------|---------------------------|--------------------|----------------------|------------
supplejack_harvester | DigitalNZ    | github.com | ../           | app    | 1.14                      | 6.1                | 0.35                 | 7.59

richardmatthewsdev:
> Would it be good to install `codeclimate_diff` so that it can be run? I was curious if there were any regressions

Yeah that's a good idea @isabel-anastasiadis-boost! I'll add it

app/frontend/js/TestTransformation.js
```diff
@@ -0,0 +1,46 @@
+import {EditorState} from "@codemirror/state"
+import {EditorView, basicSetup} from "codemirror"
+import {json} from "@codemirror/lang-json"
+
+document.addEventListener('DOMContentLoaded', function () {
+
+
+  function sendData(form) {
```
isabel-anastasiadis-boost, line 8: This is quite a big function, would it be easier to read if it was split up a bit?  It seems to do more than just send data

app/frontend/js/editor.js
```diff
@@ -2,13 +2,52 @@ import {EditorState} from "@codemirror/state"
...
+    parent: document.body,
+  });
+
+  document.querySelector('#extraction-result-viewer').append(extraction_result_viewer_editor.dom)
+}
+
+
+// Transformation Raw Data Record Viewer
+
+const raw_data_record_viewer = document.querySelector("#raw-data-record-viewer");
```
isabel-anastasiadis-boost, line 24: This code looks a bit repeated - could it be extracted to a shared function somehow?

app/frontend/stylesheets/blocks/_record-view.scss
```diff
@@ -0,0 +1,17 @@
+.record-view {
+  @include modifier('extraction-result') {
+    #extraction-result-viewer {
```
isabel-anastasiadis-boost, line 3: This isn't very BEM-y.  Is that deliberate?

motizuki (approved):

app/models/job.rb
```diff
@@ -67,4 +67,8 @@ def duration_seconds
   def extraction_folder_size_in_bytes
     `du #{extraction_folder} | cut -f1`.to_i
   end
+
+  def name
```
isabel-anastasiadis-boost, line 71: Why is the job name a date?  Do you want doc comments like the other methods?

app/views/content_partners/show.html.erb
```diff
@@ -28,5 +31,33 @@
 
   <%= render 'shared/pagination_below_table', items: @extraction_definitions %>
 <% else %>
-  <p>There are currently no extraction definitions for this content partner.</p>
+  <p>There are currently no extraction definitions for <%= @content_partner.name_in_database %>.</p>
+<% end %>
+
+<h2 class="float-start">Transformations</h2>
```
isabel-anastasiadis-boost, line 37: Would it be good to have Transformations and Extraction Definitions in their own partials?  Might make it easier to follow what is going on.  Just an idea.

app/frontend/js/TestTransformation.js
richardmatthewsdev: Yeah this is a copy of another test feature that we have. I think we will revisit these when we have a javascript framework in the mix.

app/models/job.rb
richardmatthewsdev: The date is the only identifying information we have at the moment about the job and we show the date it was run on the front end :)

app/views/content_partners/show.html.erb
richardmatthewsdev: Yeah that sounds good :)

isabel-anastasiadis-boost (approved):
Looks good Richard!

app/models/job.rb
isabel-anastasiadis-boost: Is 'name' a misnomer though?  Just wondering if it is actually a name or not.  And if it is a formatted date, is that view logic, or model logic?  Up to you, it just seemed a bit odd to me :)

app/frontend/stylesheets/blocks/_record-view.scss
isabel-anastasiadis-boost: Got it!

app/frontend/js/TestTransformation.js
isabel-anastasiadis-boost: Ahh, it could be that this is one of the code duplication things that was identified too then.

app/models/job.rb
richardmatthewsdev: That's fair :). I actually made this method so that I could get the name for the job in a rails form helper but I am not using that anymore so I will get rid of this method

app/frontend/js/TestTransformation.js
isabel-anastasiadis-boost: Could be good to at least not have the duplication if it is mostly a copy paste.  Then it is easier to keep improving later

paul-mesnilgrente (approved):
LGTM :) as promised

PR #5, opened 2023-05-03
ADD ATTRIBUTES: As an Open Source User, I want to be able to add an attribute to my transformation, so that I get all of the information I need.

richardmatthewsdev:
**Acceptance Criteria**
- User can add attributes to their transformation
- User can edit or delete an attribute
- User can run transformation with their attribute(s)
- User can preview attributes


app/controllers/fields_controller.rb
```diff
@@ -0,0 +1,44 @@
+# frozen_string_literal: true
+
+class FieldsController < ApplicationController
```
motizuki, line 3: Do we need to cater for errors/failures here?
```
if @field.save
 nice
else
 show errors?
end

File: app/frontend/js/features/AppDetailsSlice.js [hunk 2]
```diff
@@ -0,0 +1,41 @@
...
+
+    const response = request
+      .post(
+        `/content_partners/${contentPartnerId}/transformation_definitions/${transformationDefinitionId}/fields/run`,
+        {
+          record: record,
+          fields: fields,
+        }
+      )
+      .then((response) => {
```
motizuki, line 18: should we add a catch() if the post fail?

app/frontend/js/features/FieldsSlice.js
```diff
@@ -0,0 +1,111 @@
...
+        {
+          field: {
+            content_partner_id: contentPartnerId,
+            transformation_definition_id: transformationDefinitionId,
+            name: name,
+            block: block,
+          },
+        }
+      )
+      .then(function (response) {
```
motizuki, line 28: same as above

app/supplejack/transformation/field_execution.rb
```diff
@@ -0,0 +1,34 @@
...
+      WebMock.enable! unless Rails.env.test?
+      begin
+        block = ->(record) { eval(@field.block) }
+
+        @value = block.call(api_record)
+      rescue Exception => e
+        @error = e
+      end
+
+      WebMock.disable! unless Rails.env.test?
```
motizuki, line 27: hmm, I wonder when we `WebMock.enable!` if it will apply across everything, not only within the code we are running, which maybe could have side effects?

it might be good to add the `WebMock.disable` in a `ensure` block so we can ensure it always run?

motizuki (approved):
nice job @richardmatthewsdev @paul-mesnilgrente !

Just a few comments around the WebMock and some error handling

danielaboost (approved):

app/supplejack/transformation/field_execution.rb
richardmatthewsdev: Thanks for your comment here bro. I think there is a risk of the webmock disable interfering with other running jobs. If that becomes an issue we will need to look at another way of preventing HTTP calls in the transformation :)

PR #7, opened 2023-05-14
RUN A HARVEST: As an Open Source User, I want to be able to run an end-to-end harvest, so that I can aggregate metadata.

richardmatthewsdev:
**Acceptance Criteria**
- User can create a harvest
- User can run a harvest
- User can see information about the progress of a harvest
- Harvest withstands record failure
- User can edit transformation without impacting running harvests

**Notes**
- When creating a harvest, you could select a transformation and then the app will create a copy of it to be used so that the transformation being used by the harvest is locked and the harvester editing the visible one won't impact the running one.


lun-boost (approved):
Nice work @richardmatthewsdev and @paul-mesnilgrente, thanks for letting me review it. Looks really good to me based on my understanding of the codebase!

One thing that I would suggest is to break the pull request down into smaller pull requests so it helps a junior dev like myself to understand the code to learn from it and provide valuable feedback ❤️‍🔥

richardmatthewsdev:
> Nice work @richardmatthewsdev and @paul-mesnilgrente, thanks for letting me review it. Looks really good to me based on my understanding of the codebase!
> 
> One thing that I would suggest is to break the pull request down into smaller pull requests so it helps a junior dev like myself to understand the code to learn from it and provide valuable feedback ❤️‍🔥

Thanks for your feedback @lun-boost! I'm glad you enjoyed reviewing the code. Yes I am hoping our pull requests will get smaller in the future :D.

app/controllers/harvest_jobs_controller.rb
```diff
@@ -0,0 +1,52 @@
...
+
+    redirect_to content_partner_harvest_definition_harvest_job_path(@content_partner, @harvest_definition, @harvest_job)
+  end
+
+  def cancel
+    if @harvest_job.cancelled!
+      @harvest_job.extraction_job.cancelled!
+      flash.notice = 'Harvest job cancelled successfully'
+    else
+      flash.notice 'There was an issue cancelling the harvest job'
```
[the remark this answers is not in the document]
richardmatthewsdev: Yes! Thanks @lun-boost :)

spec/factories/transformation_job.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :transformation_job do
+    status { 'queued' }
+    page { 1 }
+
+    # before(:create) do |transformation_job|
+    #   binding.pry
```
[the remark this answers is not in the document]
richardmatthewsdev: Be gone! Spawn of darkness 🔪

app/supplejack/transformation/field_execution.rb
```diff
@@ -15,16 +15,16 @@ def initialize(field)
     # rubocop:disable Security/Eval
     # rubocop:disable Lint/RescueException
     def execute(api_record)
-      WebMock.enable! unless Rails.env.test?
+      # WebMock.enable! unless Rails.env.test?
```
motizuki, line 18: do we want to keep this comment?

richardmatthewsdev: Yeah this broke unfortunately like you thought it might :(.

motizuki (approved):

app/supplejack/transformation/field_execution.rb
motizuki: :'(

app/models/concerns/job.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+module Job
+  extend ActiveSupport::Concern
+
+  STATUSES = %w[queued cancelled running completed errored].freeze
```
motizuki, line 6: Im not sure if this is still a thing in rails... but I remmeber if the items in the array change its order it will cause some wierd issues because its been stored in the db as its index position.

so I'd either add a spec to ensure it has the correct order, or update it so use a hash:
https://naturaily.com/blog/ruby-on-rails-enum

richardmatthewsdev: nice! I'll make it a hash. Thanks bro :)

PR #8, opened 2023-05-21
EXTRACT ENRICHMENT DATA: As an open-source user, I want to enrich the data I have previously extracted with data from additional sources so that I can build a more complete data set.

richardmatthewsdev:
**Acceptance Criteria**
- user can retrieve data from an existing destination
- user can fetch additional data based on their existing data set
- user can explore extracted enrichment data

**Notes**
- Fetch transformed records from the API
- Fetch additional data from new sources 
- Save data on disk to be transformed


app/frontend/js/TestTransformationRecordSelector.js
```diff
@@ -0,0 +1,19 @@
+import { bindTestForm } from './utils/test-form';
+import { EditorState } from "@codemirror/state";
+import { EditorView, basicSetup } from "codemirror";
+import { json } from "@codemirror/lang-json";
+
+bindTestForm('test', 'js-test-transformation-record-selector', (response, _alertClass) => {
```
motizuki, line 6: keen to learn more about this pattern

app/views/extraction_definitions/_enrichment_form.html.erb
```diff
@@ -0,0 +1,51 @@
+<%= vertical_form_with model: [@content_partner, @extraction_definition], class: { row: true } do |form| %>
```
hapiben, line 1: Would be good to pass the instance vars as `locals`.

app/models/extraction_definition.rb
```diff
@@ -31,18 +36,26 @@ class ExtractionDefinition < ApplicationRecord
...
-  validate :total_selector_format
+
+  # Harvest related validation
+  validates :format, presence: true, inclusion: { in: %w[JSON HTML XML OAI] },            if: -> { harvest? }
+  validates :base_url, presence: true, format: { with: URI::DEFAULT_PARSER.make_regexp }, if: -> { harvest? }
+  validate :total_selector_format,                                                        if: -> { harvest? }
 
   # pagination fields
-  validates :pagination_type, presence: true, inclusion: { in: %w[item page] }
+  validates :pagination_type, presence: true, inclusion: { in: %w[item page] },           if: -> { harvest? }
```
hapiben, line 47: I think you can do something like this.

```
with_options if: :harvest?
  # validations without `if`s.
end

File: app/models/extraction_definition.rb [hunk 4]
```diff
@@ -31,18 +36,26 @@ class ExtractionDefinition < ApplicationRecord
...
   validates :page, numericality: { only_integer: true }
   validates :per_page, numericality: { only_integer: true }
-
   validates :total_selector, presence: true
 
+  # Enrichment related validation
+
+  validates :destination_id, presence: true,                                              if: -> { enrichment? }
+  validates :source_id, presence: true,                                                   if: -> { enrichment? }
+  validates :enrichment_url, presence: true,                                              if: -> { enrichment? }
```
hapiben, line 57:
```suggestion
  with_options presence: true, if: :harvest? do
    validates :destination_id
    validates :source_id
    validates :enrichment_url
  end
```

hapiben (approved):
Looking good @richardmatthewsdev. Just a few comments.

richardmatthewsdev:
> Looking good @richardmatthewsdev. Just a few comments.

Thanks for your feedback @hapiben ! I will make those changes :D

paul-mesnilgrente (approved):

PR #12, opened 2023-06-02
USER ACCOUNTS: As an Open Source user, I want to use a secure harvesting tool, so that my systems are safe. 

richardmatthewsdev:
**Acceptance Criteria**

- User can create an account with 2FA.
- User can reset their password.
- Admin can manage accounts.
- User can edit their account details.

**Notes**
Do we need a recaptcha?
Requires Design for user sign up flow.


motizuki (approved):
Nice work!, it would be nice to have this 2FA user flow extracted as a devise plugin!

PR #13, opened 2023-06-20
EXPAND EXTRACTION FORMATS: As an open-source user, I want to be able to run different types of extractions, so that we can try out more and bigger harvests.

richardmatthewsdev:
**Acceptance Criteria**
- XML format is supported 
- Tokenised pagination for JSON & XML is supported

**Risks**
- We annoy another content partner with too many API calls while testing

**Tests**
- JSON - [iNaturalist](https://manager.digitalnz.org/parsers/5bece6e9297b1f057bf25090/edit)
- XML - [Trove PIM](https://manager.digitalnz.org/parsers/59261800297b1fd4fa000004/edit)

**Notes**
- This will eventually allow us to try big a ol' [NLNZcat](https://manager.digitalnz.org/parsers/61a04909497b9b000ae64bb3/edit) harvest (XML token pagination / OAI)


richardmatthewsdev:
> I think it all looks good. There's one thing we should think about though is getting rid of all the:
> 
> ```
> if format == 'XML'
>   ...
> else if format == 'HTML'
>   ...
> ```
> 
> I feel like we should use factories and abstract objects to make clear what the intention is. Instead of using directly Nokogiri, or JSON to parse things. It would make things easier to read.

Awesome thanks @paul-mesnilgrente! Yeah I agree, it's all ready not feeling great with the different format checking so will be good to have a nicer way of handling that.

motizuki (approved):

PR #17, opened 2023-07-11
MOVE THROUGH PREVIEW RECORDS: As Ting, I want to be able to navigate through my extracted records, so that I can see how my transformations work on other records

paul-mesnilgrente:
[MOVE THROUGH PREVIEW RECORDS: As Ting, I want to be able to navigate through my extracted records, so that I can see how my transformations work on other records](https://www.pivotaltracker.com/story/show/185477527)

STORY
=====

**Acceptance Criteria**
- Transformation is able to be moved forward and backward through extracted records/pages
- Ability to jump directly to a record/page

**Risks**
- ?

**Notes**
- [Figma Link](https://www.figma.com/file/UL5rFpFs9woyHNBdspYu3C/SJ-%7C-Proof-of-Concept?type=design&mode=design)
- Another way harvesters change the focus is adjusting the record_selector eg //items[./license="cc-by"]  to test how transformation works on specific rights scenarios. This, and swapping to a different extract is achievable by editing the transformation definition.

**Impacts**
- Very important, currently can only test 1 record from the whole extraction, which might be a dull record.

### DIFF
--- a/app/controllers/concerns/redux_state.rb
+++ b/app/controllers/concerns/redux_state.rb
@@ -25,16 +25,7 @@ module ReduxState
   end
 
   def raw_record_slice
-    records = @transformation_definition.records
-    record_number = (params[:record] || 1).to_i
-    {
-      page: (params[:page] || 1).to_i,
-      record: record_number,
-      totalPages: @transformation_definition.extraction_job.documents.total_pages,
-      totalRecords: records.length,
-      format: @transformation_definition.extraction_job.format,
-      body: records[record_number]
-    }
+    RawRecordSlice.new(@transformation_definition, params[:page], params[:record]).call
   end
 
   def app_details_slice
--- a/app/controllers/concerns/redux_state.rb
+++ b/app/controllers/concerns/redux_state.rb
@@ -30,8 +30,8 @@ module ReduxState
     {
       page: (params[:page] || 1).to_i,
       record: record_number,
-      total_pages: @transformation_definition.extraction_job.documents.total_pages,
-      total_records: records.length,
+      totalPages: @transformation_definition.extraction_job.documents.total_pages,
+      totalRecords: records.length,
       format: @transformation_definition.extraction_job.format,
       body: records[record_number]
     }
--- a/app/controllers/concerns/redux_state.rb
+++ b/app/controllers/concerns/redux_state.rb
@@ -6,7 +6,7 @@ module ReduxState
   def transformation_app_state
     {
       entities: {
-        fields: fields_slice, appDetails: app_details_slice
+        fields: fields_slice, rawRecord: raw_record_slice, appDetails: app_details_slice
       },
       ui: {
         fields: ui_fields_slice, appDetails: ui_app_details_slice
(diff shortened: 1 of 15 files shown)


app/controllers/concerns/redux_state.rb
```diff
@@ -0,0 +1,73 @@
+# frozen_string_literal: true
+
+module ReduxState
```
richardmatthewsdev, line 3: Thanks for doing this bro! Good to clean up the transformation definitions controller.

app/frontend/js/apps/TransformationApp/JumpTo.jsx
```diff
@@ -0,0 +1,158 @@
...
+                value={inputPage}
+                onChange={(e) => setInputPage(parseInt(e.target.value))}
+              />
+              <span className="input-group-text">/ {totalPages}</span>
+            </div>
+          </div>
+          <div className="col-auto">
+            <button
+              className="btn btn-primary"
+              disabled={
```
richardmatthewsdev, line 86: Maybe put this logic in a clearly named method so you can see disabled={method} rather than all these conditions.

app/frontend/js/apps/TransformationApp/JumpTo.jsx
```diff
@@ -0,0 +1,158 @@
...
+                />
+                <span className="input-group-text"> / {totalRecords}</span>
+              </div>
+            </div>
+          </div>
+          <div className="col-auto">
+            <button
+              className="btn btn-primary"
+              disabled={
+                isFetching ||
```
richardmatthewsdev, line 121: Then you could use it here too :D

richardmatthewsdev (approved):
Looking good bro! :)

app/frontend/js/features/RawRecordSlice.js
```diff
@@ -0,0 +1,52 @@
...
+  "fields/askNewRawRecordStatus",
+  async (payload) => {
+    const { transformationDefinitionId, page, record } = payload;
+
+    const response = request
+      .get(
+        `/transformation_definitions/${transformationDefinitionId}/raw_records`,
+        { params: { page: page, record: record } }
+      )
+      .then(function (response) {
```
motizuki, line 14: do we need to a catch in case something goes wild?

motizuki (approved):
nice work guys!

app/frontend/js/apps/TransformationApp/JumpTo.jsx
paul-mesnilgrente: All done :)

app/frontend/js/apps/TransformationApp/JumpTo.jsx
paul-mesnilgrente: Done :)

app/frontend/js/features/RawRecordSlice.js
paul-mesnilgrente: I added a `.rejected` case for when the response fails :)

PR #21, opened 2023-07-30
REJECT AND DELETE FUNCTIONALITY: As Tim H, I want to be able to specify conditions to delete and skip records while harvesting

richardmatthewsdev:
**Acceptance Criteria**
- Records can be skipped based on specified metadata criteria (reject_if equivalent)
- Records can be deleted (status: deleted) based on specified metadata criteria (delete_if equivalent)

**Risks**
- Deletion feels like a risky action

**Notes**

**Tests**
- How can we test for success?

**Impacts**
- Where useful add other assumptions or expectations around outcomes?

### DIFF
--- a/app/controllers/raw_records_controller.rb
+++ b/app/controllers/raw_records_controller.rb
@@ -5,7 +5,7 @@ class RawRecordsController < ApplicationController
   before_action :find_transformation_definition, only: [:index]
 
   def index
-    record = @transformation_definition.records(params[:page].to_i)[params['record'].to_i]
+    record = @transformation_definition.records(params[:page].to_i)[params['record'].to_i - 1]
 
     transformation = Transformation::Execution.new([record], @fields, @reject_conditions, @delete_conditions).call.first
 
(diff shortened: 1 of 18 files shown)


motizuki (approved):

app/controllers/raw_records_controller.rb
```diff
@@ -1,9 +1,31 @@
 # frozen_string_literal: true
 
 class RawRecordsController < ApplicationController
+  before_action :find_fields, only: [:index]
+  before_action :find_transformation_definition, only: [:index]
+
   def index
-    transformation_definition = TransformationDefinition.find(params[:transformation_definition_id])
+    record = @transformation_definition.records(params[:page].to_i)[params['record'].to_i - 1]
```
motizuki, line 8: do we need a spec for this?

richardmatthewsdev: I don't think so bro, it's just accounting for the difference between page numbers and array indexes :)

PR #31, opened 2023-08-30
SJ ENRICHMENT SCOPING: As Dan, I want enrichments to run over the right records so that the only active records that have the required metadata get enriched (and deleted records stay deleted)

paul-mesnilgrente:
[SJ ENRICHMENT SCOPING: As Dan, I want enrichments to run over the right records so that the only active records that have the required metadata get enriched (and deleted records stay deleted)](https://www.pivotaltracker.com/story/show/185538724)

STORY
=====

**Acceptance Criteria**
- `status:deleted` records are not included in enrichments
- Ability to specify required conditions/metadata to scope the enrichments based on existing API record
- Enrichments run over the right number of records (ie only run once per record)

**Risks**
- If an enrichment picks up an intentionally deleted record and it ends up active, that is bad news.

**Notes**
- Consider if this should be a change on the API or the harvester 
- Gus did something similar with current SJ so might be worth talking to him 
- Newshub enrichment was being blocked because its looking at an old deleted record that the landing_url doesnt work, so I cant build the enrich transformation
- Are there other parts of SJ that need to be similarly restricted?
- Richard said this about a recent FigShare harvest "It looks like that enrichment did enrich each record twice"
https://harvester.staging.digitalnz.org/pipelines/4/harvest_definitions/14/harvest_jobs/34
- [This enrichment](https://harvester.staging.digitalnz.org/pipelines/15/harvest_definitions/28/extraction_definitions/36/edit?referrer_id=15) uses a piece of metadata from the original record to construct the loop up url. But not all records have that metadata, so we need to scope the enrichment to only those records that have the required metadata.
- The second AC essentially needs access to the API record, which is something that is needed in enrichments in order to replicate the same functionality as existing SJ

**Tests**
- https://harvester.staging.digitalnz.org/content_sources/42/extraction_definitions/21/edit


richardmatthewsdev (approved):
Looking good bro! Thanks for fixing the delete records appearing bug :D.

PR #36, opened 2023-09-07
IMPLEMENT SJ CARD INFORMATION CHANGES: As Dan I want cards in the new SJ tool to show useful information, so that harvesters have the details and context they need at all levels of the harvesting workflow

paul-mesnilgrente:
[IMPLEMENT SJ CARD INFORMATION CHANGES: As Dan I want cards in the new SJ tool to show useful information, so that harvesters have the details and context they need at all levels of the harvesting workflow](https://www.pivotaltracker.com/story/show/185839774)

STORY
=====

**Acceptance Criteria**
- Information shown on SJ cards gives the harvester confidence they are working with the right components (as per notes)

**Risks**
- too much info
- information not technically available

[DESIGNS](https://www.figma.com/file/UL5rFpFs9woyHNBdspYu3C/SJ-%7C-Proof-of-Concept?type=design&node-id=3883%3A6582&mode=design&t=VYF6RIYc1rfes19X-1)

**Notes**

Extraction definition
- Last edited (by:), pages/records, referenced 

Transformation definition
- Last edited (by:), fields, referenced 

Pipelines
- Last edited (by:), Number of harvests/enrichments


app/controllers/concerns/error_handling.rb
```diff
@@ -0,0 +1,28 @@
...
+  def render_bad_request(resource)
+    render_bad_request_json(resource)
+  end
+
+  def render500
+    format.html { render500_html }
+    format.json { render500_json }
+  end
+
+  def render500_html
```
isabel-anastasiadis-boost, line 17: Should this be here if it isn't used yet?

app/controllers/concerns/error_handling.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+module ErrorHandling
+  extend ActiveSupport::Concern
+
+  protected
+
+  def render_bad_request(resource)
+    render_bad_request_json(resource)
```
isabel-anastasiadis-boost, line 9: where is this method defined?

app/controllers/concerns/last_edited_by.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  protected
+
+  # Updates the attribute "last_edited_by" with the current user
+  # The resource can be given as a parameter this method uses the
+  # `last_edited_by_resources` method that can be defined in the
+  # controller
+  #
+  # @params resource: a model having the last_edited_by attribute
+  def update_last_edited_by(resource = nil)
```
isabel-anastasiadis-boost, line 14: This method does seem pretty complicated. Couldn't you just pass in the resource?

app/controllers/application_controller.rb
```diff
@@ -1,10 +1,12 @@
 # frozen_string_literal: true
 
 class ApplicationController < ActionController::Base
+  include ErrorHandling
   include UserAuthorization
   include DeviseOverrides
   include TransformationReduxState
   include ExtractionReduxState
+  include LastEditedBy
```
isabel-anastasiadis-boost, line 9: I just wonder if this is actually going to be used by EVERY controller, and if not, might be a bit clearer if you include the concern in each controller that needs it

app/controllers/pipelines_controller.rb
```diff
@@ -66,6 +66,6 @@ def assign_sort_by
   end
 
   def pipeline_params
-    params.require(:pipeline).permit(:name, :description)
+    merge_last_edited_by(params.require(:pipeline).permit(:name, :description))
```
isabel-anastasiadis-boost, line 69: This could just be me, and it is pretty minor, so feel free to disregard if you don't agree, but having it all in one statement makes it a bit harder to reason about.

I think it would be clearer if you did:
```
safe_params = ....
merge_last_edited_by(safe_params)
```

app/controllers/parameters_controller.rb
```diff
@@ -2,29 +2,43 @@
...
+      render500
+    end
   end
 
   def update
-    @parameter.update(parameter_params)
-
-    render json: @parameter
+    if @parameter.update(parameter_params)
+      update_last_edited_by
```
isabel-anastasiadis-boost, line 20: Sorry if this is a dumb question, but if the last updated by id is added to the params, why do you need to call this method after it is updated?

app/frontend/stylesheets/bootstrap/_card.scss
```diff
@@ -4,6 +4,16 @@
   width: 100%;
   position: relative;
 
+  .card-title {
```
isabel-anastasiadis-boost, line 7: I'm assuming this is overrides for Bootstrap and that's why it isn't BEM? :) If it can be BEM that would be good

app/views/pipelines/_card.html.erb
```diff
@@ -1,19 +1,32 @@
...
    run_sample_extraction_path ||= ''
-   run_full_extraction_path ||= '' %>
+   run_full_extraction_path ||= ''
+   type = definition.class.name.gsub(/Definition$/, '').downcase
+   position = type == 'extraction' ? 'first' : 'last' %>
 
 <div class="card">
   <div class='card-body'>
-    <h5 class="card-title"><%= definition.send("#{type}_definition").name %></h5>
+    <h5 class="card-title"><%= definition.name %></h5>
```
isabel-anastasiadis-boost, line 9: Ah I see this is our own controlled classes. Should use BEM then :)

app/views/pipelines/_card.html.erb
```diff
@@ -1,19 +1,32 @@
...
+    <h6 class="card-subtitle">
+      <%= last_edited_by(definition) %>
```
isabel-anastasiadis-boost, line 11: There's a fair bit of logic here, would it be good to put it in a partial and test it with a view test?

public/site.webmanifest
```diff
@@ -0,0 +1,19 @@
+{
+    "name": "",
```
isabel-anastasiadis-boost, line 2: Are these deliberately empty?

app/views/pipelines/_card.html.erb
isabel-anastasiadis-boost: Or maybe just have a view test for this partial

app/views/pipelines/show.html.erb
```diff
@@ -87,15 +91,17 @@
       <% end %>
     <% end %>
```
isabel-anastasiadis-boost, line 92: Probably not to fix in this story but this view is massive! Might be good to look for an opportunity to extract some partials

isabel-anastasiadis-boost (approved):
Looks good Paul, sorry I left a lot of comments but feel free to use your judgement about how to respond to them.

One final thing is it is a pretty large story, so I'd suggest next time breaking it up a bit. But otherwise it's looking good, and I liked your context and the tests :D

app/controllers/concerns/error_handling.rb
paul-mesnilgrente: That's indeed true, I didn't even think about remove the `format.html`, now done.

app/controllers/concerns/error_handling.rb
paul-mesnilgrente: That was a left over, I removed it now :)

app/controllers/concerns/last_edited_by.rb
paul-mesnilgrente: It was indeed a bit too complicated for what it was doing, I simplified it :)

app/controllers/pipelines_controller.rb
paul-mesnilgrente: I don't have a big opinion, I just went your way which seems to make things clearer

app/controllers/parameters_controller.rb
paul-mesnilgrente: That's now clearer with the refactoring done

app/frontend/stylesheets/bootstrap/_card.scss
paul-mesnilgrente: It is overwriting bootstrap styles

app/views/pipelines/_card.html.erb
paul-mesnilgrente: Do you think? I don't want to recreate the bootstrap cards. I just want to make the bootstrap cards fit the design.

app/views/pipelines/_card.html.erb
paul-mesnilgrente: Actually I'll make this helpers

public/site.webmanifest
paul-mesnilgrente: Those files were auto-generated, I didn't think about filling those. Now done.

app/controllers/concerns/last_edited_by.rb
isabel-anastasiadis-boost: Nice!

app/controllers/pipelines_controller.rb
isabel-anastasiadis-boost: Cool!

app/frontend/stylesheets/bootstrap/_card.scss
isabel-anastasiadis-boost: Ah I see, cool yes I think it is fine if we're overriding the overriding bootstrap card styling globally.  I think I'm getting confused because you've got elements included lower down in this file.  They don't look like bootstrap overrides. Maybe not for this story, but they should probably go in a BEM block somewhere else then.

app/frontend/stylesheets/bootstrap/_card.scss
```diff
@@ -4,6 +4,16 @@
...
+    font-size: 1.125rem;
+  }
+
+  .card-subtitle {
+    color: $pompeii-ash;
```
isabel-anastasiadis-boost, line 12: I think subtitle colour is configurable with Bootstrap variables

app/controllers/parameters_controller.rb
isabel-anastasiadis-boost: I think I was more wondering why sometimes you are adding the last updater in the params, and sometimes by manually calling `update_last_edited_by`.  Do you need both strategies? Or am I misunderstanding the difference between them?

app/views/pipelines/_card.html.erb
isabel-anastasiadis-boost: That's fair, I forgot a bit how bootstrap / foundation works.  All good!

isabel-anastasiadis-boost (approved):

app/controllers/application_controller.rb
paul-mesnilgrente: Done

PR #38, opened 2023-09-21
SHARED DEFINITIONS (II) CLONING: As Dan I want to be able to use and edit an existing definition safely, so that I don't accidentally change existing pipelines.

richardmatthewsdev:
**Acceptance Criteria**
- User can create an exact copy of a shared definition that can be edited independently
- When user 'Creates new definition from shared' definition name is copied and prefaced with '[CLONE]'
- The new definition is no longer shared and has no association with old definition

**Notes**
- Consider how we can prep this functionality to be applied to entire pipelines

[Design concepts](https://www.figma.com/file/UL5rFpFs9woyHNBdspYu3C/SJ-%7C-Proof-of-Concept?type=design&node-id=4032%3A9047&mode=design&t=5BTlZM3xov8fzmNT-1)

### DIFF
--- a/app/controllers/extraction_definitions_controller.rb
+++ b/app/controllers/extraction_definitions_controller.rb
@@ -68,18 +68,13 @@ class ExtractionDefinitionsController < ApplicationController
   end
 
   def clone
-    cloned_extraction_definition = @extraction_definition.clone(
-      @pipeline,
-      @harvest_definition,
-      extraction_definition_params['name']
-    )
+    cloned_extraction_definition = @extraction_definition.clone(@pipeline, @harvest_definition,
+                                                                extraction_definition_params['name'])
 
     if cloned_extraction_definition
-      redirect_to edit_pipeline_harvest_definition_extraction_definition_path(
-        @pipeline,
-        @harvest_definition,
-        cloned_extraction_definition
-      ), notice: t('.success')
+      flash.notice = t('.success')
+      redirect_to edit_pipeline_harvest_definition_extraction_definition_path(@pipeline, @harvest_definition,
+                                                                              cloned_extraction_definition)
     else
       flash.alert = t('.failure')
       redirect_to pipeline_path(@pipeline)
@@ -126,10 +121,8 @@ class ExtractionDefinitionsController < ApplicationController
 
   def extraction_definition_params
     safe_params = params.require(:extraction_definition).permit(
-      :pipeline_id,
-      :name, :format, :base_url, :throttle, :page, :per_page,
-      :total_selector, :kind, :destination_id, :source_id, :enrichment_url,
-      :paginated
+      :pipeline_id, :name, :format, :base_url, :throttle, :page, :per_page,
+      :total_selector, :kind, :destination_id, :source_id, :enrichment_url, :paginated
     )
     merge_last_edited_by(safe_params)
   end
--- a/app/controllers/extraction_definitions_controller.rb
+++ b/app/controllers/extraction_definitions_controller.rb
@@ -68,10 +68,11 @@ class ExtractionDefinitionsController < ApplicationController
   end
 
   def clone
-    cloned_extraction_definition = @extraction_definition.clone(@pipeline, @harvest_definition,
+    cloned_extraction_definition = @extraction_definition.clone(@pipeline,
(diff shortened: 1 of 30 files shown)


motizuki (approved):

app/controllers/fields_controller.rb
```diff
@@ -57,7 +57,7 @@ def find_transformation_definition
   end
 
   def find_fields
-    @fields = @transformation_definition.fields.where(id: params['fields'])
+    @fields = @transformation_definition.fields.where(id: params['fields']).order(created_at: :desc)
```
isabel-anastasiadis-boost, line 60: What is this change for? Does it need a unit test?

app/models/extraction_definition.rb
```diff
@@ -64,4 +67,17 @@ def to_h
   def shared?
     harvest_definitions.count > 1
   end
+
+  def clone(pipeline, name)
+    cloned_extraction_definition = ExtractionDefinition.new(dup.attributes.merge(name:, pipeline:))
+
+    requests.each do |request|
+      cloned_request = request.dup
```
isabel-anastasiadis-boost, line 75: Might be not understanding, but I wonder if delegating what it means to clone a request would be better put on the request model (if that is a model)

app/models/harvest_definition.rb
```diff
@@ -40,4 +40,8 @@ def to_h
       }
     }
   end
+
+  def clone(pipeline)
+    HarvestDefinition.new(dup.attributes.merge(pipeline:))
```
isabel-anastasiadis-boost, line 45: Maybe dumb question, but couldn't work out where `dup` came from. What is it?

isabel-anastasiadis-boost: Also not sure what `dup.attributes.merge(pipeline:)` does.  Just letting you know in case you expect it to be obvious, so you can decide if it needs a comment.

isabel-anastasiadis-boost (approved):
I think it looks pretty good Richard. I have a few questions, more just saying there's stuff that isn't immediately obvious to me, so you can choose if you think that should be commented.

app/controllers/fields_controller.rb
richardmatthewsdev: This is for a change that was lost on the front end that I fixed here while I was here. I will add a test so it doesn't go missing again :)

app/models/harvest_definition.rb
richardmatthewsdev: Not a dumb question! `dup` is an ActiveRecord method which returns a copy of the record but it doesn't have the id or the created_at. It also doesn't bring across the associations. 

`dup.attributes` returns a hash of new attributes that I want to assign to the `HarvestDefinition` and then I merge the `pipeline` argument into it to assign the `HarvestDefinition` to the new pipeline, rather than the old one that was in the duplicates attributes. 

https://apidock.com/rails/ActiveRecord/Core/dup

app/models/extraction_definition.rb
richardmatthewsdev: Yeah I was wondering about this too. It's probably fine here but I don't feel too strongly about it.

app/controllers/extraction_definitions_controller.rb
```diff
@@ -89,6 +102,15 @@ def update_redirect_path
     end
   end
 
+  def successful_clone_path(clone)
```
paul-mesnilgrente, line 105: This method should return a path so that it follows its name. You'll also be able to do 

`redirect_to successful_clone_path(clone), notice: t('.success')` in one line :)

config/routes.rb
```diff
@@ -31,6 +35,10 @@
           post :test_enrichment_extraction
         end
 
+        member do
+          post :clone
+        end
```
paul-mesnilgrente, line 40: `post :clone, on: :member` ?

paul-mesnilgrente (approved):

app/controllers/extraction_definitions_controller.rb
richardmatthewsdev: Nice! Thanks @paul-mesnilgrente

config/routes.rb
richardmatthewsdev: 🎫

PR #40, opened 2023-09-25
RUBOCOP

paul-mesnilgrente:
the "db" and "spec" folder were getting ignored so I made a step forward to include them again and fix them.


richardmatthewsdev (approved):
Thanks @paul-mesnilgrente! I'm really glad we have rubocop in the pipeline now too.

PR #41, opened 2023-09-26
SHARED DEFINITIONS (III ) DELETIONS: As Dan I want to be able to delete a definition from my pipeline so that I can easily replace it with another one.

richardmatthewsdev:
**Acceptance Criteria**
- User can delete/remove a definition from a pipeline
 - If a definition is used in more than one pipeline: "Remove shared definition"
 - If a definition is only used in this pipeline: "Delete definition"
with a warning "Are you sure? This cannot be undone"
- Definition is not dependent on its original pipeline
- User can delete a block from a pipeline (deletion logic applies as above to definitions in block).
- Clean up any existing orphaned/lost/dud definitions

**Risks**
- Permanently deleting something that we want back

**Notes**
[Design concepts](https://www.figma.com/file/UL5rFpFs9woyHNBdspYu3C/SJ-%7C-Proof-of-Concept?type=design&node-id=4032%3A9046&mode=design&t=VYF6RIYc1rfes19X-1)

**Impacts**
- Really looking forward to being able to delete/remove stuff!!

### DIFF
--- a/app/helpers/pipelines_helper.rb
+++ b/app/helpers/pipelines_helper.rb
@@ -15,6 +15,16 @@ module PipelinesHelper
     "Edit #{type}"
   end
 
+  def definition_delete_text(definition, type)
+    if definition.shared?
+      "<p>This will <strong>NOT</strong> delete the #{type} definition '#{definition.name}'
+          as it is shared with another pipeline.</p>"
+    else
+      "<p>This <strong>WILL</strong> delete the #{type} definition '#{definition.name}'
+          as it is <strong>NOT</strong> shared with another pipeline.</p>"
+    end
+  end
+
   private
 
   def extraction_definition_help_text(definition, type)
--- a/app/helpers/pipelines_helper.rb
+++ b/app/helpers/pipelines_helper.rb
@@ -16,15 +16,11 @@ module PipelinesHelper
   end
 
   def definition_delete_text(definition, type)
-    # rubocop:disable Rails/OutputSafety
     if definition.shared?
-      "<p>This will <strong>NOT</strong> delete the #{type} definition '#{definition.name}'
-          as it is shared with another pipeline.</p>".html_safe
+      "This will NOT delete the #{type} definition '#{definition.name}' as it is shared with another pipeline."
     else
-      "<p>This <strong>WILL</strong> delete the #{type} definition '#{definition.name}
-          as it is <strong>NOT</strong> shared with another pipeline.</p>".html_safe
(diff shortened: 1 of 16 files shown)


app/helpers/pipelines_helper.rb
```diff
@@ -15,6 +15,18 @@ def definition_edit_text(definition, type)
     "Edit #{type}"
   end
 
+  def definition_delete_text(definition, type)
+    # rubocop:disable Rails/OutputSafety
+    if definition.shared?
+      "<p>This will <strong>NOT</strong> delete the #{type} definition '#{definition.name}'
```
hapiben, line 21: Can this be a `content_tag` or `tag` so you can remove the `rubocop:disable`?

app/helpers/pipelines_helper.rb
```diff
@@ -15,6 +15,18 @@ def definition_edit_text(definition, type)
...
+          as it is shared with another pipeline.</p>".html_safe
+    else
+      "<p>This <strong>WILL</strong> delete the #{type} definition '#{definition.name}
+          as it is <strong>NOT</strong> shared with another pipeline.</p>".html_safe
+    end
+    # rubocop:enable Rails/OutputSafety
+  end
```
paul-mesnilgrente, line 28: I am not feeling comfortable with this html_safe as `definition.name` is user inputs...

Could you use `tag.p do` and `tag.strong` etc. to avoid the html_safe? not sure if it will help.

hapiben (approved):

app/helpers/pipelines_helper.rb
richardmatthewsdev: Yeah I was trying to do that aye, do you know if there is a way to output a content tag that has a strong tag nested inside of it?

I almost had it working with some spans but it wasn't ideal.

config/locales/en.yml
```diff
@@ -73,6 +73,9 @@ en:
     update:
       success: Harvest Definition updated successfully
       failure: There was an issue updating your Harvest Definition
+    destroy:
+      success: "%{kind} deleted successfully"
+      failure: "There was an issue deleting your %{kind}"
```
paul-mesnilgrente, line 78: I love this line :joy:

paul-mesnilgrente (approved):

app/helpers/pipelines_helper.rb
richardmatthewsdev: Fair enough, I will drop the strong as it's not really critical and use a content tag :)

PR #45, opened 2023-10-16
IMPROVE ENRICHMENT EXTRACTION INTERFACE: As Dan, I want the Enrichment Extraction interface updated, so that the workflow is intuitive and I can better customise my extraction

richardmatthewsdev:
**Acceptance Criteria**
- The user is easily able to navigate the process and understand the 2 concepts involved (API record vs Enrichment URL)
- Preview presents response in a useful way
- Preview is helpful with errors and isn't a dead end if the first record doesn't meet requirements.
- Conditions allow enrichment to skip based on record metadata (eg skip category:Audio)
- Target (destination) where enrichment is pulling from is intuitive (and automatic where possible)

**Details**
- API record selection
 - source_id - could be auto-populated from  pipeline (but might need to deal with being empty?)
 - Conditions - allow enrichment to skip based on record metadata (eg skip category:Audio)
- Enrichment details 
 - URL construction (involving base_url + record metadata)
 - Format (html, XML etc)
 - Headers (eg optional authentication)
 - Throttle - only applied to external requests (with a valid URL)
 - HTTP verb (get/post)

**Risks**
- What problems could this cause (mitigations? notify dev-all-dnz?)
- Is there some extra investment that could add value/learning/QA

**Notes**
- This [Enrichment](https://manager.digitalnz.org/parsers/54f3cfa374544802c9000004/edit) needs Auth details in a header
- The transformation error feedback about the "wrong type" would be really useful for when this goes wrong `record['dc_identifier']` and needs a `.first.to_s`. Or when Dan uses a symbol instead of a string.
- Also the Preview functionality that shows the actual URL being called will be super useful.
- BUG: The enrichment URL evaluation where it checks the curly brackets is brittle. Things like `"#{record['hello']}/iamcool/#{record['goodbye']}"`  will likely not work. Also if the URL is just `record['landing_url']` (without quotes or curly brackets) it will fail

**Notes**

Gus has said that we need the ability to enrich from an API source, this could be a POST request with some headers and slugs based on information in the record. 
Harvest From PDF

https://github.com/DigitalNZ/supplejack_harvester/assets/1848911/68d48418-7f5e-4985-b42e-c13255c15ca1


motizuki (approved):
nice improvement!

lun-boost (approved):
Win win @richardmatthewsdev !! Thanks for including the video 💇‍♀️

PR #47, opened 2023-10-30
SJ SLOWNESS WITH LARGE EXTRACTIONS: As Tim I want SJ to handle large extractions efficiently, so that my workflow is smooth and my experience is good

richardmatthewsdev:
**Acceptance Criteria**
- SJ handles large extractions efficiently and operators do not get frustrated with slow page loads

**Solution Notes**

What we have decided to do here is allow the harvest operator to split the resulting documents from an Extraction Definition. This means instead of a single large XML file full of records we end up with lots of little ones that can then flow into the normal use case. Currently we only support XML but we will look to change this in the future.


app/frontend/entrypoints/application.js
```diff
@@ -36,7 +36,13 @@ import "/js/TestDestination";
 import "/js/Tooltips";
 import "/js/Toasts";
 import "/js/CollapseScroll";
-import "/js/formModals";
+
```
isabel-anastasiadis-boost, line 39: is the extra space deliberate?

app/frontend/js/modals/createModal.js
```diff
@@ -0,0 +1,11 @@
+import { Modal } from "bootstrap";
+
+const createModal = document.getElementById("create-modal");
+
+if (createModal) {
```
isabel-anastasiadis-boost, line 5: This code looks very similar to addHarvestModal.js.  Does codeclimate pick it up?  You could consolidate it

app/frontend/js/modals/transformationDefinitionSettingsModal.js
```diff
@@ -57,38 +37,46 @@ if (transformationDefinitionSettingsForms) {
...
-      } else if (format == "XML") {
-        result = xmlFormat(result, {
-          indentation: "  ",
-          lineSeparator: "\n",
-        });
-      }
-
-      editor(`#js-record-selector-result-${id}`, format, true, result);
-      tooltip.disable();
+      if (result == "") {
```
isabel-anastasiadis-boost, line 40: There are a lot of nested if else, and this file is pretty large and complicated. Could it be broken up to make it more readable?

isabel-anastasiadis-boost: I'm guessing some of it could be shared functionality like those other simpler modals?

app/frontend/js/modals/updateExtractionDefinitionModal.js
```diff
@@ -0,0 +1,73 @@
+import { Modal } from "bootstrap";
+import { each } from "lodash";
+
+const updateExtractionDefinitionModal = document.getElementById(
+  "update-extraction-definition-modal"
+);
+
+if (updateExtractionDefinitionModal) {
```
isabel-anastasiadis-boost, line 8: Same with this file, could it be broken up a bit?

config/brakeman.ignore
```diff
@@ -111,6 +111,29 @@
...
+        "type": "method",
+        "class": "SplitWorker",
+        "method": "perform"
+      },
+      "user_input": "ExtractionJob.find(extraction_job_id).extraction_folder",
+      "confidence": "Medium",
+      "cwe_id": [
+        22
+      ],
+      "note": ""
```
isabel-anastasiadis-boost, line 135: Should you put a note in here to say why it isn't a concern?

app/frontend/entrypoints/application.js
richardmatthewsdev: Yeah I did it to group those modals together, it's probably not necessary though :)

app/frontend/js/modals/createModal.js
richardmatthewsdev: Yeah that sounds good :). I actually haven't run codeclimate for a bit :(

app/frontend/js/modals/transformationDefinitionSettingsModal.js
richardmatthewsdev: Iv split this one up :)

app/frontend/js/modals/updateExtractionDefinitionModal.js
richardmatthewsdev: I think this one is OK for now, it's not as crazy as the other one

config/brakeman.ignore
richardmatthewsdev: Good idea! 💃

motizuki (approved):

app/frontend/js/modals/transformationDefinitionSettingsModal.js
```diff
@@ -0,0 +1,153 @@
...
+  "js-transformation-definition-form"
+);
+
+if (transformationDefinitionSettingsForms) {
+  each(transformationDefinitionSettingsForms, (form) => {
+    initializeTransformationPreview(form);
+  });
+}
+
+function initializeTransformationPreview(form) {
```
isabel-anastasiadis-boost, line 18: I think this is better, but it is still a bit hard to understand because the functions are chained, and the names don't necessarily make it clear what is happening.

I wonder if it would be clearer if you could break it down into sequential steps with good names, make functions for those, and then call all the functions sequentially.

Eg. Don't know if this is right, but from reading the code, maybe something like:

1. Find elements to update
2. Extract data
3. Update elements with data
4. Bind form (not really sure what this does)

Does that make sense?

app/frontend/js/modals/updateExtractionDefinitionModal.js
```diff
@@ -0,0 +1,63 @@
...
+    extractionDefinitionFormat.addEventListener("change", (event) => {
+      toggleSplitElements(event.target.value);
+    });
+
+    splitDropdown.addEventListener("change", () => {
+      toggleSplitElements(extractionDefinitionFormat.value);
+    });
+  }
+
+  function toggleSplitElements(format) {
```
isabel-anastasiadis-boost, line 27: Might be good to add a comment saying what this is doing. It isn't immediately clear to me what this means.

app/frontend/js/modals/updateExtractionDefinitionModal.js
isabel-anastasiadis-boost: If you find breaking up the other one into steps works, might still be good to do the same here.

isabel-anastasiadis-boost (approved):
Looks good!

PR #48, opened 2023-11-14
[BUG] RECORD SELECTOR CRASH: As Dan I want the harvester to handle incorrect record_selector's gracefully, so that the user is informed and able to correct the problem

richardmatthewsdev:
When adding a record selector for XML or HTML, if you get it wrong it can cause an error in Nokogiri which can crash the page when it is trying to pull records from the document.

### Steps to reproduce
1. In a transformation record selector add something that wont work like `/no:namespace`

### Expected
It should save and not work, but not crash

### Actual
Instantly "We're sorry, but something went wrong"


lun-boost (approved):
⭐

paul-mesnilgrente (approved):

PR #51, opened 2023-12-07
SJ RUN SET NUMBER TRANSFORMATIONS: As Dan I want to be able to limit the size of my harvest, so that I can test out samples of my larger harvests

richardmatthewsdev:
### Steps to reproduce
1. Run a harvest, with an existing extraction, for a set number of pages 

### Expected
Transform will only run over set pages

### Actual
The Transform will run over all records rather than set pages


paul-mesnilgrente (approved):
don't forget the commit message conventions :)

PR #53, opened 2023-12-20
EXTRACTION STOP CONDITIONS (Part II): As Dan, I want a way to define when an extraction should stop so that the harvest does not run forever

richardmatthewsdev:
**Acceptance Criteria**
- User can define stop conditions when building the extraction

**Notes**
- Should it stop if the record_selector finds no records?

Screenshot 2024-01-09 at 11 06 00 AM


app/models/stop_condition.rb
```diff
@@ -0,0 +1,27 @@
...
+  belongs_to :extraction_definition
+
+  # rubocop:disable Lint/UnusedBlockArgument
+  # rubocop:disable Security/Eval
+  def evaluate(document)
+    block = ->(response) { eval(content) }
+
+    block.call(document)
+  rescue StandardError => e
+    e
```
motizuki, line 13: do we need to do something with `e`?

motizuki (approved):

app/models/stop_condition.rb
richardmatthewsdev: Not at the moment. We don't have any error reporting yet but once we do we will populate the report with errors :).

paul-mesnilgrente (approved):

PR #55, opened 2024-01-12
2 SJ Bugs

richardmatthewsdev:
### Steps to reproduce
1. Update the settings to use the cancelled extraction job. 
2. Save the modal.

### Expected
You should be able to transform a cancelled extraction.

### Actual
the page does not load fully and gets stuck on 
`Waiting for extraction to run. Please wait a minute and refresh the page.`

----------------------------------------------------------------------------------

--------------------------------------------------------------

### Steps to reproduce
1. Notice the "destination" before pressing "Edit" on a schedule

### Expected
The page should auto populate with the same "destination" from the previous screen

### Actual
It seems to default to staging (even though it was previously Prod)

### DIFF
--- a/.github/workflows/code_quality_score.yml
+++ b/.github/workflows/code_quality_score.yml
@@ -6,4 +6,4 @@ on:
 jobs:
   code_quality_score:
     name: Code quality
-    uses: DigitalNZ/code_quality_score/.github/workflows/calculate-code-quality-score.yml@main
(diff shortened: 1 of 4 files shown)


lun-boost (approved):
Niceee! 👯

eoin-boost (approved):
looks good, nice to have the code quality in too!

PR #58, opened 2024-01-22
SLOW LOADING PIPELINES: As Tim, I want my pipelines to load in a timely manner, so that I can have a good flow and get my work done efficiently

paul-mesnilgrente:
[SLOW LOADING PIPELINES: As Tim, I want my pipelines to load in a timely manner, so that I can have a good flow and get my work done efficiently](https://www.pivotaltracker.com/story/show/186497741)

STORY
=====

**Acceptance Criteria**
- Pipeline page loads efficiently (ideally snappily, under 2 seconds)
- or draft stories to fix if the problem is complicated

**Risks**
- ?

**Notes**
- It feels like the some pages (eg pipeline page) are unnecessarily loading the full extractions or some large unnecessary data... Or is that data needed for some reason?

**Tests**
Loading this pipeline took 4 seconds!
clicking into the extraction definition was snappy
but clicking back to the pipeline then took 19 seconds!
initial load to 5.4 seconds
clicking into extraction definition didnt load!! (I gave up after 1 minute, maybe a separate bug)
click back to pipeline 4.6 seconds

### DIFF
--- a/app/controllers/pipelines_controller.rb
+++ b/app/controllers/pipelines_controller.rb
@@ -79,12 +79,8 @@ class PipelinesController < ApplicationController
 
   def assign_show_variables
     @harvest_definition = @pipeline.harvest_definitions.find(&:harvest?) || HarvestDefinition.new(pipeline: @pipeline)
-
+    @extraction_jobs = @harvest_definition.extraction_definition&.extraction_jobs&.order(created_at: :desc)
     @enrichment_definition = HarvestDefinition.new(pipeline: @pipeline)
-
-    return if @harvest_definition&.extraction_definition.blank?
-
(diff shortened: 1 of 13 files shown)


app/controllers/pipelines_controller.rb
```diff
@@ -61,8 +63,25 @@ def clone
 
   private
 
+  def assign_show_pipeline
+    # binding.pry
```
richardmatthewsdev, line 67: # binding.pry

richardmatthewsdev (approved):
Looking good bro! Thanks for doing this :)

app/controllers/pipelines_controller.rb
paul-mesnilgrente: Removed

dgpangan (approved):

PR #62, opened 2024-03-08
Extract data from attached documents

richardmatthewsdev:
Supplejack Users can extract raw text from a PDF document.


Dockerfile
```diff
@@ -3,7 +3,7 @@ FROM ruby:3.2.2-alpine3.18
 WORKDIR /app
 
 ARG BUILD_PACKAGES="build-base curl-dev git"
-ARG DEV_PACKAGES="bash mysql-client mariadb-dev yaml-dev zlib-dev nodejs yarn libxml2 libxml2-dev libxslt libxslt-dev gmp-dev"
+ARG DEV_PACKAGES="bash mysql-client mariadb-dev yaml-dev zlib-dev nodejs yarn libxml2 libxml2-dev libxslt libxslt-dev gmp-dev openjdk8-jre"
```
richardmatthewsdev, line 6: Java is needed for Yomu file -> text conversion

Dockerfile
```diff
@@ -34,6 +34,8 @@ ARG RAILS_ENV="production"
 ENV RAILS_ENV=$RAILS_ENV
 ARG RAILS_MASTER_KEY
 ENV RAILS_MASTER_KEY=$RAILS_MASTER_KEY
+ARG RAILS_SECRET_KEY_BASE
```
richardmatthewsdev, line 37: Secret key base needs to be in the project secrets

config/credentials.yml.enc
richardmatthewsdev, line 1: This file is unusable for an open source user as the master key is not available (which it shouldn't be)

danielaboost:
Included in https://github.com/DigitalNZ/supplejack_harvester/pull/66

PR #71, opened 2024-07-11
INDEXING DROPPING RECORDS: As Dan I want all records to be reliably indexed after being updated, so that a harvests changes are accurately reflected in the API

paul-mesnilgrente:
[INDEXING DROPPING RECORDS: As Dan I want all records to be reliably indexed after being updated, so that a harvests changes are accurately reflected in the API](https://www.pivotaltracker.com/story/show/187794086)

STORY
=====

**Acceptance Criteria**
- All records updated in a harvest are indexed correctly in solr

**Risks**
- Is this wider than just new SJ enrichments?!

**Problem**
- After a harvest and enrichment:
 - all records get their metadata updated correctly
 - a handful of records fail to be indexed and still appear in search for previous metadata
 - those records id's were listed in the array to be indexed
 - those records are able to be indexed (eg if you do them manually they work)
- First test 12 / 287 missed
- Second test 2 / 476 missed
- Third test 108 / 287 missed

**Dan's Theory**
- New SJ is not updating the "Active source ids"
`"Active source ids [...]`
This is used to ensure the indexing doesn't start picking up records until they have been finished updating/harvesting/enriching. Implemented in this story https://www.pivotaltracker.com/story/show/181472631

**Notes**
- There has been a lot of work lately in this area of our system:
 - https://www.pivotaltracker.com/story/show/182068934
 - Solr and Mongo upgrades
 - https://www.pivotaltracker.com/story/show/187598418
- Does not appear to be happening in old SJ

**Tests**
- Run an enrichment where the enrichment transformation adds bulk value to all records
- Query that source_id + the bulk added value
- Re-run the enrichment with a modified value
- Re-check the query and see a few left over in solr, but note that their metadata has been updated (shows new value, but is returned in solr query for old value)
- eg this shows 108 records that failed to be indexed :(

**Impacts**
- Indexing has to be reliable!

WHAT WAS DONE
==============

- Updated the API to allow it to store source_ids being harvested
- Updated the harvester to tell the API when a source_id is being harvested and it stops being harvested.


app/supplejack/api/harvester/record.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
```
richardmatthewsdev, line 1: Nice! Good idea with these bro :)

motizuki (approved):

app/supplejack/api/utils/notice_harvesting_to_api.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+module Api
+  module Utils
+    class NoticeHarvestingToApi
```
richardmatthewsdev, line 5: Just a little thing but I feel like the name of this class is a bit odd, it doesn't really read grammatically. Maybe just `NotifyHarvesting`? So the module would be `Api::Utils::NotifyHarvesting`

richardmatthewsdev (approved):
Thanks for doing this bro!

app/supplejack/api/utils/notice_harvesting_to_api.rb
paul-mesnilgrente: I wasn't satisfied with it either. I'll change it to NotifyHarvesting, that's better

PR #73, opened 2024-07-16
Harvest text from PDF images

richardmatthewsdev:
As an Agency, I want the text in my PDF image to be able to be harvested by Supplejack, so that information about my legislation can be accurately used on the NZLegislation website. 

**Acceptance Criteria**
- Text in PDF images can be harvested
- Viewers of the data are able to see if the data has been converted from a PDF or Image

**Notes**
- Harvester must be able to determine dynamically whether it is dealing with an image or text file


spec/sidekiq/text_extraction_worker_spec.rb
```diff
@@ -9,40 +9,92 @@
...
+  
+        it 'converts a PDF into raw text' do
+          extracted_files = Dir.glob("#{extraction_job.extraction_folder}/*").select { |e| File.file? e }
+    
+          expect(extracted_files.count).to eq 1
+  
+          TextExtractionWorker.new.perform(extraction_job.id)
   
-        expect(extracted_files.count).to eq 1
+          extracted_files = Dir.glob("#{extraction_job.extraction_folder}/*").select { |e| File.file? e }
```
eoin-boost, line 25: question: probably a fundamental ruby question but do you need to redefine the variable in order to reevaluate this? Not sure if this is a pass by reference vs pass by value question

app/sidekiq/text_extraction_worker.rb
```diff
@@ -3,23 +3,51 @@
...
+      text = ocr_pdf(file)
+      process = 'Extracted from PDF using OCRmyPDF'
+    end
+
+    { text:, process: }
+  end
+
+  def ocr_pdf(file)
+    base_file_name = File.basename(file, File.extname(file))
+    `ocrmypdf \
```
eoin-boost, line 43: question: guessing there was no gem available that would work?

eoin-boost (approved):
great feature thanks for this @richardmatthewsdev

motizuki (approved):

PR #74, opened 2024-07-22
Schema

richardmatthewsdev:
Add the ability to define a set of fields in a schema that can be used in a Transformation


app/frontend/js/apps/SchemaApp/SchemaApp.jsx
```diff
@@ -0,0 +1,35 @@
...
+
+// actions from state
+
+import { selectSchemaFieldIds } from '~/js/features/SchemaApp/SchemaFieldsSlice';
+
+const SchemaApp = () => {
+  const schemaFieldIds = useSelector(selectSchemaFieldIds);
+
+  return (
+    <>
```
danielaboost, line 17: is this empty tag a redux thing or an accident? 🤔

app/frontend/js/apps/SchemaApp/components/SchemaField.jsx
```diff
@@ -0,0 +1,290 @@
...
+        schemaFieldId: id,
+        schemaId: appDetails.schema.id
+      })
+    );
+
+    setSchemaFieldValue('');
+  };
+
+  return (
+    <>
```
danielaboost, line 119: Same question as above for this one!

app/helpers/schema_fields_helper.rb
danielaboost, line 1: @richardmatthewsdev did this get left behind?

app/views/schemas/edit.html.erb
danielaboost, line 1: Is this also leftovers?

app/views/schemas/new.html.erb
danielaboost, line 1: Is this also left over or does it interact with JSX in some way I don't know about 🤔

danielaboost (approved):

danielaboost:
@richardmatthewsdev approved! Thank you so so much for your hard work - after having a play around with this feature its already making building pipelines faster 😁 

A small suggestion - it would be awesome if the defined values added to a schema field display in alphabetical order in both the list on the schema page, and the select when its on a model (I noticed yesterday with 46 agencies it would make this data a bit easier to work with).

I know you've spent a lot of time on this, so we can do it as a different story! GREAT WORK ✨

app/frontend/js/apps/SchemaApp/components/SchemaField.jsx
```diff
@@ -0,0 +1,290 @@
...
+  setActiveSchemaField,
+} from "~/js/features/SchemaApp/UiSchemaFieldsSlice";
+
+
+import { addSchemaFieldValue } from '~/js/features/SchemaApp/SchemaFieldValuesSlice';
+
+import Button from "react-bootstrap/Button";
+import Modal from "react-bootstrap/Modal";
+
+import SchemaFieldValue from "~/js/apps/SchemaApp/components/SchemaFieldValue";
```
eoin-boost, line 23: could a shorter relative reference be used?

app/frontend/js/apps/SchemaApp/components/SchemaFieldNavigationPanel.jsx
```diff
@@ -0,0 +1,70 @@
+import React from "react";
+import { useSelector, useDispatch } from "react-redux";
+import { selectAllSchemaFields } from "~/js/features/SchemaApp/SchemaFieldsSlice";
+import SchemaFieldNavigationListItem from "./SchemaFieldNavigationListItem";
+import AddSchemaField from "~/js/apps/SchemaApp/components/AddSchemaField";
```
eoin-boost, line 5: relative ref could be good again?

app/views/schemas/new.html.erb
eoin-boost, line 1: assuming we can get rid of this file?

eoin-boost (approved):
great work @richardmatthewsdev

app/frontend/js/apps/SchemaApp/SchemaApp.jsx
richardmatthewsdev: This is a way of returning multiple HTML elements in one component without having to wrap them in an empty div :)

app/helpers/schema_fields_helper.rb
richardmatthewsdev: Thanks!

app/views/schemas/edit.html.erb
richardmatthewsdev: Thanks!

PR #75, opened 2024-07-24
Fix broken deletion for extraction jobs

danielaboost:
Harvest Definitions & Extraction Definitions cannot currently be deleted via the UI due to destroy restrictions.

Have chatted with Richard and updated the harvest jobs to be able to be deleted in the deletion cascade

### DIFF
--- a/app/models/extraction_definition.rb
+++ b/app/models/extraction_definition.rb
@@ -12,7 +12,7 @@ class ExtractionDefinition < ApplicationRecord
   belongs_to :last_edited_by, class_name: 'User', optional: true
 
   # the before_destroy needs to be here (before any other dependent: :destroy statements)
-  before_destroy :remove_dependent_objects, prepend: true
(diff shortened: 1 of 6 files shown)


app/models/extraction_definition.rb
```diff
@@ -80,4 +83,17 @@ def clone(pipeline, name)
 
     cloned_extraction_definition
   end
+
+  def destroy_associated_definitions
+    # Remove all associated harvest reports
+    extraction_jobs.each do |extraction_job|
+      next if extraction_job.harvest_job.blank?
+
+      extraction_job.harvest_job.harvest_report.destroy!
```
eoin-boost, line 92: question: would it be possible to have delete cascade of associated harvest report of a harvest job? Perhaps we don't want this behaviour for some reason

motizuki (approved):

eoin-boost (approved):
thanks for this @danielaboost

PR #76, opened 2024-07-25
Past jobs can still be viewed when an enrichment definition or harvest definition is removed

danielaboost:
Reports were dependent on their enrichment definitions and harvest definitions, and they were being destroyed if the definition was deleted. This meant the reports tab was breaking as the report no longer existed.

We're fixed this by making reports an independent object, which copies the pipeline name that it belongs to (the only piece of data that was missing once the definition had been deleted)

### DIFF
--- a/app/views/pipeline_jobs/index.html.erb
+++ b/app/views/pipeline_jobs/index.html.erb
@@ -67,7 +67,7 @@
                ) %>
 
               <tr>
-                <td><%= report.harvest_job.name %></td>
+                <td><%= report.definition_name || report.harvest_job.name %></td>
                 <td>
                   <% if pipeline_job.schedule.present? %>
                     Schedule
@@ -90,7 +90,7 @@
                 <td>
                   <i class="bi bi-three-dots-vertical table__control" data-bs-toggle='dropdown'></i>
                   <ul class="dropdown-menu dropdown-menu-end">
-                    <% if report.harvest_job.extraction_job_id.present? %>
+                    <% if report.harvest_job && report.harvest_job.extraction_job_id.present? %>
                       <li>
                         <%= link_to pipeline_harvest_definition_extraction_definition_extraction_job_path(
                               @pipeline.id,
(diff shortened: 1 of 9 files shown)


motizuki (approved):

app/views/pipeline_jobs/index.html.erb
```diff
@@ -67,7 +67,7 @@
                ) %>
 
               <tr>
-                <td><%= report.harvest_job.name %></td>
+                <td><%= report.definition_name || report.harvest_job.name %></td>
```
eoin-boost, line 70: question: should the definition name always be available and therefore we could remove the ||?

spec/models/extraction_definition_spec.rb
```diff
@@ -204,8 +204,8 @@
         expect { subject.destroy }.to change(HarvestJob, :count).by(-1)
       end
 
-      it "destroys the harvest reports" do
-        expect { subject.destroy }.to change(HarvestReport, :count).by(-1)
+      it "does not destroy the harvest reports" do
```
eoin-boost, line 207: suggestion: came across this yesterday inspired by your TDD talk which may be of interest if not applicable here. https://rspec.info/features/3-12/rspec-core/subject/one-liner-syntax/

spec/models/harvest_report_spec.rb
```diff
@@ -12,7 +12,13 @@
   let(:harvest_job)        { create(:harvest_job, harvest_definition:, pipeline_job:) }
 
   describe 'associations' do
-    it { is_expected.to belong_to(:pipeline_job) }
+    it { is_expected.to belong_to(:pipeline_job).optional }
+  end
+
+  describe '#initialize' do
+    it 'has the same name as its harvest definition' do
+      expect(subject.definition_name).to eq harvest_definition.name
```
eoin-boost, line 20: suggestion: [this](https://rspec.info/features/3-12/rspec-core/subject/one-liner-syntax/) could be of interest again here 😄

eoin-boost (approved):
looks good thank you @danielaboost

PR #77, opened 2024-07-25
Gm/retry extractions

motizuki:
Extraction process is retried when it fails.

- Setup a default `Retriable` config: 
```
Retriable.configure do |c|
  c.tries = 10
  c.base_interval = 3
  c.multiplier = 2
end
```

- Retry the `AbstractExtraction#extract` when an exception occur

### DIFF
--- a/config/initializers/retriable.rb
+++ b/config/initializers/retriable.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+Retriable.configure do |c|
+  c.tries = 10
+  c.base_interval = 3
+  c.multiplier = 2
+end
--- a/config/initializers/retriable.rb
+++ b/config/initializers/retriable.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 Retriable.configure do |c|
-  c.tries = 10
-  c.base_interval = 3
-  c.multiplier = 2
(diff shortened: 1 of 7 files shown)


spec/spec_helper.rb
```diff
@@ -98,3 +98,8 @@
   # as the one that triggered the failure.
   Kernel.srand config.seed
 end
+
+require 'retriable'
+Retriable.configure do |c|
+  c.tries = 2
+end
```
hapiben, line 105: **suggestion:** Move this to `spec/support`. Uncomment this line https://github.com/DigitalNZ/supplejack_harvester/blob/b80146ff56083d3642b1f6faca40be6f983a6f4b/spec/rails_helper.rb#L28 and remove this one. https://github.com/DigitalNZ/supplejack_harvester/blob/b80146ff56083d3642b1f6faca40be6f983a6f4b/spec/rails_helper.rb#L11

hapiben (approved):

config/initializers/retriable.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+Retriable.configure do |c|
+  c.tries = 10
```
richardmatthewsdev, line 4: Awesome thanks bro, could we make this configurable with an environment variable as well?

richardmatthewsdev (approved):
🥳

PR #78, opened 2024-08-01
Concurrent Enrichment Extractions

danielaboost:
**Acceptance Criteria**
- Enrichments can be configure to be run concurrently
- Any other improvement to enrichments are welcome
- consider what settings might be useful to customise / safeguard this process

**Notes**
- If you look at the Act, Bills, Regulation pipelines enrichments, they are a good example of enrichments we would like to speed up
- not for every use case

**What we've done**
- Created a new worker to run the job async when the setting is selected for running the pipeline. The existing logic has been moved into a concern so the same logic can be used both for async runs and non-async runs

Screenshot 2024-08-07 at 11 07 40 AM


danielaboost:
> ## Code quality score
> Uh oh! The code quality got worse for this PR! Better take a look!! 🚨
> 
> Ruby file count	Similarity score (flay)	ABC complexity (flog)	Code smells (reek)	TOTALS
> base	90	6.12	5.36	17.2	28.68
> this branch	93	6.17	5.36	17.25	28.78
> difference	3	⚠️ 0.05	0.0	⚠️ 0.05	0.1

I have refactored this A LOT. At the beginning the CQ increase was a lot worse, and i've managed to get it to here which is far better. The remaining reeks are things that I don't necessarily agree with, such as moving the EnrichmentExtractionProcess concern private methods into a utility class (which I don't love as I want to keep the logic near where its being used since they should only ever be used within the `process_enrichment_extraction` method).

Needless to say I am comfortable with the increase and have thought about this a lot 😁

app/supplejack/extraction/concerns/enrichment_extraction_process.rb
```diff
@@ -0,0 +1,92 @@
...
+          return unless enrichment_extraction.valid?
+
+          enrichment_extraction.extract_and_save
+          handle_harvest_job(extraction_context)
+        end
+      end
+
+      private
+
+      ExtractionContext = Struct.new(
```
richardmatthewsdev, line 23: Is this also a value object? I feel like it would be good to define it outside of this class.

app/supplejack/extraction/concerns/enrichment_extraction_process.rb
```diff
@@ -0,0 +1,92 @@
...
+      included do
+        def process_enrichment_extraction(enrichment_params)
+          parsed_params = JSON.parse(enrichment_params)
+          extraction_context = build_extraction_context(parsed_params)
+
+          enrichment_extraction = extraction_context.enrichment_extraction
+          return unless enrichment_extraction.valid?
+
+          enrichment_extraction.extract_and_save
+          handle_harvest_job(extraction_context)
```
richardmatthewsdev, line 17: maybe trigger_harvest_job? handle isn't very clear what it means

app/views/pipelines/_run_pipeline.html.erb
```diff
@@ -126,6 +126,25 @@
           </div>
         </div>
 
+         <div class='col-5'>
+          <%= form.label :run_enrichment_concurrently, class: 'form-label' do %>
+            Run Enrichment Concurrently
```
richardmatthewsdev, line 131: Little thing but I think this should be plural as you can have multiple enrichments

app/views/pipelines/_run_pipeline.html.erb
```diff
@@ -126,6 +126,25 @@
...
+
+            <span
+              data-bs-toggle="tooltip"
+              data-bs-title="Selecting this option when running your pipeline will run multiple enrichments at once"><i class="bi bi-question-circle" aria-label="helper text"></i>
```
richardmatthewsdev, line 135: This tooltip isn't correct, it's more that the extractions within an enrichment will run in parallel. Even without this option multiple enrichments will run at once :)

spec/value_objects/extraction_params_spec.rb
```diff
@@ -0,0 +1,3 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
```
richardmatthewsdev, line 3: Could we add some specs for this class?

richardmatthewsdev (approved):
Thanks @danielaboost! Very cool

tom-mcculloch (approved):
Nice work mate, cool improvement!

PR #79, opened 2024-08-01
feat: Add docker compose file

eoin-boost:
Adding docker compose to make the process of getting started using Supplejack Harvester easier.


richardmatthewsdev (approved):
Thanks for this @eoin-boost!

danielaboost (approved):

PR #80, opened 2024-08-01
ADD FIELD FOCUSED ELEMENT

danielaboost:
As a harvester I want the first action element (text box) within a supplejack field to be focused once i've added it so that I can add fields faster without having to click into the first element

**Acceptance Criteria**
- When a 'supplejack field' is added the first actionable element (right now the field name) is focused for the user

**What we've done**
Made the first field when adding a field (the 'name' field) focused on add as shown below

Screenshot 2024-08-02 at 11 30 45 AM


richardmatthewsdev (approved):
💃

richardmatthewsdev:
Do we want to add this to the extraction app as well? So when a user adds a parameter it has the same behaviour.

eoin-boost (approved):
thanks for this @danielaboost

PR #82, opened 2024-08-22
[TIMEBOX: 16 HRS] Extract from Javascript pages

richardmatthewsdev:
As an agency, I want my javascript based website to be able to be harvested, so that my secondary legislation can be seen on the NZLegislation website. 

**Acceptance Criteria**

- Supplejack can extract text from a web page that renders it with Javascript. 
- If we can't do it, propose an alternative.

**Notes**
- 8 Hour Check In
- Document anything that makes sense within the story 

This site renders the results with javascript and we can't harvest them https://environment.govt.nz/acts-and-regulations/national-direction/.
Here's another one: https://www.ird.govt.nz/income-tax/income-tax-for-businesses-and-organisations/types-of-business-expenses/claiming-vehicle-expenses

https://www.stats.govt.nz/methods/measuring-child-poverty-concepts-and-definitions/


danielaboost (approved):

app/supplejack/extraction/javascript_request.rb
```diff
@@ -0,0 +1,56 @@
...
+    def get
+      begin
+        @driver.get(full_url)
+        @document = document(200)
+      rescue
+        @document = document(500)
+      end
+      
+      # Quit after assigning the document so that the browser process is stopped
+      @driver.quit
```
motizuki, line 22: I wonder if this could be in a `ensure` block

https://ruby-doc.org/docs/ruby-doc-bundle/UsersGuide/rg/ensure.html

> The ensure clause does just this. ensure goes after the last rescue clause and contains a chunk of code that will always be executed as the block terminates. It doesn't matter if the block exits normally, if it raises and rescues an exception, or if it is terminated by an uncaught exception---the ensure block will get run.

motizuki (approved):

app/supplejack/extraction/javascript_request.rb
richardmatthewsdev: That sounds good! I will give it a go :)

PR #83, opened 2024-08-26
feat: Optimize performance for large-scale harvest/enrichment

motizuki:
- Implement folder grouping for extraction files to mitigate performance issues caused by large numbers of files in a single directory.

- Previously, when enriching a collection of 20k items, performance degraded significantly as the extraction folder grew in size due to slow file operations.

- Files are now grouped into subdirectories in batches of 100, significantly improving performance for large datasets.

- ⚠️ BREAKING CHANGE: Existing extractions must be recreated to be compatible with the new directory structure.


app/sidekiq/split_worker.rb
```diff
@@ -21,6 +21,10 @@ def create_document(records, saved_response)
       params: saved_response['params'], request_headers: saved_response['request_headers'],
       status: saved_response['status'], response_headers: saved_response['response_headers'],
       body: "<?xml version=\"1.0\"?><root><records>#{records.map(&:to_xml).join}</records></root>"
-    ).save("#{@extraction_folder}/#{name_str}__-__#{page_str}.json")
+    ).save("#{@extraction_folder}/#{folder_number}/#{name_str}__-__#{page_str}.json")
+  end
+
+  def folder_number
```
richardmatthewsdev, line 27: I think you can move this to the parent class

app/sidekiq/text_extraction_worker.rb
```diff
@@ -19,11 +19,15 @@ def create_document(extracted_text, filename, process)
       params: saved_response['params'], request_headers: saved_response['request_headers'],
       status: saved_response['status'], response_headers: saved_response['response_headers'],
       body: { text: extracted_text, process: }.to_json
-    ).save("#{@extraction_folder}/#{filename}")
+    ).save("#{@extraction_folder}/#{folder_number}/#{filename}")
   end
 
   private
 
+  def folder_number
```
richardmatthewsdev, line 27: Since these 2 classes both inherit from the FileExtractionWorker

richardmatthewsdev (approved):
Thanks heaps for doing this bro! It looks like it is making a massive difference to the performance of the enrichments :)

app/supplejack/extraction/enrichment_extraction.rb
```diff
@@ -38,7 +38,8 @@ def headers
     def file_path
       name_str = @extraction_definition.name.parameterize(separator: '_')
       page_str = format('%09d', @page)[-9..]
-      "#{@extraction_folder}/#{name_str}__#{@record['id']}__#{page_str}.json"
+      folder_number = ((@page || 1) / Documents::DOCUMENTS_PER_FOLDER.to_f).ceil
```
paul-mesnilgrente, line 41: I see this line in a few files, could we refactor that?

paul-mesnilgrente (approved):
LGTM I just had one comment

PR #90, opened 2024-09-12
feat(specify-enrichment-fields): Allow specifying particular fields from the Enrichment to prevent returning unnecessary fields

richardmatthewsdev:
A good example is if you have a fulltext field that you do not need for the enrichment, it gets returned anyway and takes up space in the filesystem and slows down the response. Especially if it is in the merged fragment as well as an embedded fragment.


danielaboost (approved):

motizuki (approved):
NICE

PR #98, opened 2025-02-26
Stand alone enrichments

richardmatthewsdev:
As a harvest operator, I want to be able to run an enrichment that queries for records that match a certain criteria, rather than by source id, so that I can run the same process over a batch of unrelated records. 

**Acceptance Criteria**

- Pipelines are allowed to only contain enrichments
- Shared enrichment configuration is combined into appropriate pipelines

**Notes**

At the moment we have act status, act status from latest version, act latest version, bill_status, bill_status_from_latest_version, bill_latest_version, regulation_status, regulation_status_from_latest_version, regulation_latest_version which all do essentially the same thing but for different types. It would be great if we could run 1 `status` enrichment with a particular criteria that would allow us to enrich a specific set of unrelated records.


lunmanlee (approved):
Good stuff bro

motizuki (approved):

PR #113, opened 2025-07-01
Indication of activity per ETL step

boost-tim:
Add Last Updated field to pipeline jobs so we can see when the last activity on that pipeline was for harvest/enrichment/transformation.


hapiben (approved):

richardmatthewsdev (approved):
Nice! Thanks @boost-tim :)

PR #114, opened 2025-07-03
Fix timing on Enrichment jobs

boost-tim:
Accurate timing info is shown for the enrichments

### DIFF
--- a/app/views/pipeline_jobs/index.html.erb
+++ b/app/views/pipeline_jobs/index.html.erb
@@ -91,7 +91,7 @@
                   </span>
                 </td>
 
-                <% if report.harvest_job.present? && report.harvest_job&.extraction_job %>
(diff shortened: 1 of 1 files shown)


app/views/pipeline_jobs/index.html.erb
```diff
@@ -90,7 +90,13 @@
                     <%= report.status.capitalize %>
                   </span>
                 </td>
-                <td><%= pipeline_job.start_time.strftime('%H:%M %d/%m/%y') %></td>
+
+                <% if report.harvest_job&.extraction_job.present? %>
```
richardmatthewsdev, line 94: My only concern with this I think is that if you run a harvest that is using an existing extraction it would show the time that that extraction first ran. Can you check what happens in this scenario?

PR #118, opened 2025-07-14
Fix job cancellation delays by adding cancellation checks to workers

hapiben:
## Summary

  Fixes an issue where cancelled jobs continued running for minutes after users clicked "Cancel" in the UI. Jobs now stop within 10-30 seconds of cancellation instead of continuing for 2-5 minutes.

  ## Problem

  When users cancelled pipeline jobs, the UI showed "Cancelled" immediately but workers continued processing:
  - **LoadWorker** processed all remaining 100-record batches without checking cancellation status
  - **TransformationWorker** spawned new LoadWorker and DeleteWorker instances even after cancellation
  - Jobs that should stop immediately continued running until completion

  This led to:
  - Poor user experience (jobs appeared "stuck")
  - Wasted computational resources
  - Confusion about job status

  ## Solution

  Added strategic cancellation checks to key workers:

  ### LoadWorker
  - Check `@harvest_job.cancelled?` and `@harvest_job.pipeline_job.cancelled?` before each batch
  - Break out of processing loop immediately if cancelled

  ### TransformationWorker
  - Check cancellation status after transformation but before queuing child workers
  - Prevent spawning LoadWorker/DeleteWorker if job is already cancelled
  - Refactored `child_perform` method to reduce complexity (RuboCop compliance)


motizuki (approved):

richardmatthewsdev (approved):
Thanks for this bro! There is also a scenario with cancelling that doesn't currently work, can you check if this fixes it? 

Currently if I am running a job that uses an existing extraction and I cancel it, it doesn't cancel. I think your change will fix this but good to confirm :)

hapiben:
> Thanks for this bro! There is also a scenario with cancelling that doesn't currently work, can you check if this fixes it?
> 
> Currently if I am running a job that uses an existing extraction and I cancel it, it doesn't cancel. I think your change will fix this but good to confirm :)

Hi @richardmatthewsdev. I checked with Dan and we tested the cancellation together. Seems to work as expected.

richardmatthewsdev:
> > Thanks for this bro! There is also a scenario with cancelling that doesn't currently work, can you check if this fixes it?
> > Currently if I am running a job that uses an existing extraction and I cancel it, it doesn't cancel. I think your change will fix this but good to confirm :)
> 
> Hi @richardmatthewsdev. I checked with Dan and we tested the cancellation together. Seems to work as expected.

Sick! Thanks @hapiben :D

PR #122, opened 2025-07-31
Zoomed out view of harvest activity

paul-mesnilgrente:
[As Haki, I want wider visibility across harvest activity from multiple pipelines, so that I can better understand the timing and sequence of jobs, manage load and troubleshoot problems.](https://app.shortcut.com/nlnz/story/160361/zoomed-out-view-of-harvest-activity)

**Acceptance Criteria**
- Can see harvest information of more than one pipeline at a time. 
- Can see all statuses (active, completed, etc)
- Can see destination info for each harvest

**Notes**
**Richard has a concern this is going to be very slow**

- This is a daily process for harvesters to scan across overnight harvests looking for anomalies
- Needs to show all status's, including 'active' as Its sometimes good to see the list of the previous 20 harvest including any that are still active, to see where they slotted in. 
- Existing SJ also has an indication of errors (a count) on this zoomed out view. We might consider adding this later. 
- There is a semi equivalent route available https://harvester.digitalnz.org/pipelines?status=completed


richardmatthewsdev (approved):
Looking good bro! I'll be interested to know what the performance of this page is like :)

boost-tim (approved):

PR #124, opened 2025-08-25
SJ timezone inconsistency

paul-mesnilgrente:
[SJ timezone inconsistency](https://app.shortcut.com/nlnz/story/158096/sj-timezone-inconsistency)

As Tim H, I want all SJ Harvester to operate in the same timezone I do, so that things happen when I expect them to.

**Acceptance Criteria**
- All dates/times in SJ Harvester show the same timezone (preferably NZST) in the UI

**Risks**
- Could this stuff anything up behind the scenes?
- Is it risky for PCO if we change any code?
- Could there be confusions between this and some UTC AWS moving parts?

**Notes**
- Question: Will daylight savings be annoying?
- Good to do this before working on the extraction lifecycle policy.
- If harvesters work around this adding more, it will mean more work later unpicking things when we finally do fix it.
- Some API's involve date queries eg `Date.today - 1.days`. The sooner we get this right the better.
- A harvester was able to mess about with this changing an ENV[''] variable. The fact that they were messing about with this shows its still a problem.

**Steps**
- Open a day
- Find a weekly schedule
- Click 'edit' on that schedule
- See that the day its set to run, is not the day you opened.

and this
- `Time.now.zone` shows NZST in old SJ vs UTC in new SJ (this has since changed)

and this
-  the format of dates (that show up in the API after harvesting):
-- Old SJ produces - 2008-01-01T12:00:00.000Z
-- New SJ produces - 2008-01-01T00:00:00+00:00

**Impacts**
- Schedules
- Job details


Gemfile.lock
```diff
@@ -13,73 +13,73 @@ GIT
...
-      activerecord (= 7.1.5.1)
-      activesupport (= 7.1.5.1)
+    activestorage (7.1.5.2)
+      actionpack (= 7.1.5.2)
+      activejob (= 7.1.5.2)
+      activerecord (= 7.1.5.2)
+      activesupport (= 7.1.5.2)
       marcel (~> 1.0)
-    activesupport (7.1.5.1)
+    activesupport (7.1.5.2)
```
paul-mesnilgrente, line 82: Fixes some CVEs

app/models/concerns/cron_utilities.rb
```diff
@@ -17,7 +17,7 @@ def refresh_sidekiq_cron_job
   end
 
   def cron_syntax
-    "#{minute} #{hour} #{month_day} #{month} #{day_of_the_week}"
+    "#{minute} #{hour} #{month_day} #{month} #{day_of_the_week} #{Time.zone.tzinfo.name}"
```
paul-mesnilgrente, line 20: Helps the gem "fugit" interpret cron jobs in their time zones.

app/models/schedule.rb
```diff
@@ -88,7 +92,7 @@ def schedule_name(subject)
   # This is for converting 12 hour times into 24 hour times
   # so if the user has a time of 7:45pm, it becomes 19:45
   def sanitized_time
-    return DateTime.parse(time).strftime('%H:%M') if time.downcase.include?('am') || time.downcase.include?('pm')
+    return Time.zone.parse(time).strftime('%H:%M') if time.downcase.include?('am') || time.downcase.include?('pm')
```
paul-mesnilgrente, line 95: Rubocop feedback

app/supplejack/extraction/documents.rb
```diff
@@ -34,6 +34,8 @@ def total_pages
 
     def total_folders
       Dir.children(@folder).count { |f| !f.ends_with?('tmp') }
+    rescue Errno::ENOENT # folder does not exist
```
paul-mesnilgrente, line 37: A tiny fix for when you don't have all the folders.

spec/models/schedule_spec.rb
```diff
@@ -95,155 +95,163 @@
...
         }
       }
 
-      expect(schedule_map).to eq result
+      expect(schedule_map.length).to eq result.length
+
+      result.each do |date, schedules_by_time|
+        expect(schedule_map).to have_key(date)
+        expect(schedule_map[date]).to eq schedules_by_time
+      end
```
paul-mesnilgrente, line 237: Not a big fan of this loop, but it's a lot easier to debug this test!

mrGrazy (approved):
Nice one.

PR #126, opened 2025-09-03
Cloning an extraction should copy stop conditions

boost-tim:
As Tim H, I want cloned extractions to include the stop conditions from the original definition, so that a pipeline is not accidentally run without stop conditions

Steps to reproduce

Starting from a pipeline with an extraction that has some stop conditions
eg https://harvester.digitalnz.org/pipelines/182
Click the 3 dots button on the extraction and choose "Clone and edit new definition"
Current Behavior

The "first request" and "following request" tabs have details copied. (this is correct)
But the "stop conditions" tab is missing the details that should have been cloned (this is the problem)
Expected Behavior

"stop conditions" tab should have copied across the details from original extraction definition


app/models/extraction_definition.rb
```diff
@@ -78,6 +78,11 @@ def clone(pipeline, name)
       cloned_extraction_definition.requests << cloned_request
     end
 
+    stop_conditions.each do |stop_condition|
```
richardmatthewsdev, line 81: Can we have a test for this? We have other tests for the clone feature

boost-tim: @richardmatthewsdev updated the specs :)

boost-tim:
> Any tests that need to be updated?

Updated the specs now for clone :)

app/models/extraction_definition.rb
richardmatthewsdev: Thanks bro! :)

richardmatthewsdev (approved):

PR #131, opened 2025-09-18
Follow Redirects Option for Extraction Definition

boost-tim:
As Tim H, I want new SJ to work with scroll API's, so that I can successfully migrate Te Papa's harvest to new SJ

Acceptance Criteria

New SJ works with Te Papa's scroll harvest


app/supplejack/concerns/http_client.rb
```diff
@@ -3,10 +3,21 @@
 module HttpClient
   extend ActiveSupport::Concern
 
-  def connection(url, params, headers)
-    Faraday.new(url:, params:, headers:) do |f|
-      f.response :follow_redirects, limit: 5
-      f.adapter Faraday.default_adapter
+  # rubocop:disable Metrics/MethodLength
+  # rubocop:disable Style/OptionalBooleanParameter
```
boost-tim, line 7: I tried using the `follow_redirects: true` style but it didn't work, so had to use the `=`

app/models/parameter.rb
```diff
@@ -20,13 +22,28 @@ def incremental_evaluation(response_object)
...
     block = ->(response) { eval(content) }
 
-    Parameter.new(
-      name:,
-      content: block.call(response_object&.body)
-    )
+    if content.include?('headers')
+      Parameter.new(
+        name:,
+        content: block.call(OpenStruct.new(
```
richardmatthewsdev, line 34: Do we need to use this OpenStruct or can we just return the response_object?

app/supplejack/extraction/connection.rb
```diff
@@ -5,46 +5,7 @@
 # As the client itself is abstracted away from our HTTP calls
 #
```
richardmatthewsdev, line 6: I think the naming of this makes it unclear to know that it's for the follow redirects feature

app/supplejack/extraction/request.rb
```diff
@@ -4,8 +4,12 @@ module Extraction
   # Makes the actual request
   # Stores the request and response details
   class Request
-    def initialize(url:, params: {}, headers: {}, method: 'get')
-      @connection = Connection.new(url:, params:, headers:, method:)
+    def initialize(url:, params: {}, headers: {}, method: 'get', follow_redirects: true)
+      @connection = if follow_redirects
```
richardmatthewsdev, line 8: Yeah i feel like this distinction between Connection and DefaultConnection is unclear

app/supplejack/extraction/connection.rb
boost-tim: This comment is in the base connection class now, so I'll remove this here as it's confusing

app/supplejack/extraction/request.rb
boost-tim: I've made some changes here to rename these classes

app/supplejack/extraction/document_extraction.rb
```diff
@@ -10,17 +10,20 @@ def initialize(request, extraction_folder = nil, response = nil)
       @response = response
     end
 
+    # rubocop:disable Layout/LineLength
```
paul-mesnilgrente, line 13: Same comment as above, let's avoid creating exceptions to rules :D

app/supplejack/extraction/enrichment_extraction.rb
```diff
@@ -11,17 +11,20 @@ def initialize(request, record, page = 1, extraction_folder = nil)
       @extraction_folder = extraction_folder
     end
 
+    # rubocop:disable Layout/LineLength
```
paul-mesnilgrente, line 14: Same

paul-mesnilgrente (changes requested):
It's mostly looking good to me, I only don't like the rubocop exceptions in here.

app/models/parameter.rb
```diff
@@ -20,22 +20,31 @@ def incremental_evaluation(response_object)
...
     Parameter.new(
       name:,
       content: "#{content}-evaluation-error".parameterize
     )
   end
+
   # rubocop:enable Lint/UnusedBlockArgument
   # rubocop:enable Security/Eval
-
+  # rubocop:enable Metrics/MethodLength
```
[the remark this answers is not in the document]
boost-tim: Yep done

PR #132, opened 2025-09-21
Why did this Extraction stop?

chigginsboost:
This PR adds better visibility into why SJ jobs don't finish as expected. It starts to capture all the places we might encounter a stopped pipeline, then saves any relevant error information. I've created a new JobCompletionSummary model to store this information.

This is the first step, and we will most likely be updating the logging once we get feedback from the harvesters. Additionally, we'd like to update the UI so we can more easily see which specific job has failed and why. We are waiting to complete this in the upcoming [Job Management](https://app.shortcut.com/nlnz/story/161349/job-management-design-review) story.

I have focused on harvests and extractions. There is a follow up story to work on transformations. Jobs might fail due to stop conditions (which usually are set by the harvester) or errors. There are also stop conditions that are not set by the harvester, and come from the system.

I've also created a very simple dashboard to view the completion summaries:
Screenshot 2025-09-22 at 9 22 46 AM

### DIFF
--- a/app/controllers/job_completion_summary_controller.rb
+++ b/app/controllers/job_completion_summary_controller.rb
@@ -1,0 +1,25 @@
+# frozen_string_literal: true
+
+class JobCompletionSummaryController < ApplicationController
+    before_action :set_job_completion_summary, only: [:show]
+  
+    def index
+      @job_completion_summary = JobCompletionSummary.recent_errors.page(params[:page])
+  
+      if params[:error_type].present?
+        @job_completion_summary = @job_completion_summary.where(error_type: params[:error_type])
+      end
+  
+      return if params[:extraction_id].blank?
+  
+      @job_completion_summary = @job_completion_summary.where(extraction_id: params[:extraction_id])
+    end
+  
+    def show; end
+  
+    private
+  
+    def set_job_completion_summary
+      @job_completion_summary = JobCompletionSummary.find(params[:id])
+    end
+  end
--- a/app/models/job_completion_summary.rb
+++ b/app/models/job_completion_summary.rb
@@ -1,0 +1,105 @@
+# frozen_string_literal: true
+
+class JobCompletionSummary < ApplicationRecord
+    validates :extraction_id, presence: true, uniqueness: true
+    validates :extraction_name, presence: true
+    validates :error_type, presence: true
+    validates :error_details, presence: true
+    validates :error_count, presence: true, numericality: { greater_than_or_equal_to: 0 }
+  
+    scope :by_error_type, ->(type) { where(error_type: type) }
+    scope :recent, -> { order(last_error_at: :desc) }
+    scope :recent_errors, -> { recent }
+    scope :with_errors, -> { where('error_count > 0') }
+  
+    def self.group_by_error_type
+      group(:error_type).count
+    end
+  
+    def self.total_errors
+      sum(:error_count)
+    end
+  
+    def stop_condition?
+      error_type == 'stop condition'
+    end
+  
+    def system_stop_condition?
+      stop_condition? && error_details.any? { |detail| detail.dig('details', 'is_system_condition') }
+    end
+  
+    def user_stop_condition?
+      stop_condition? && !system_stop_condition?
+    end
+  
+    def self.log_error(extraction_id:, extraction_name:, message:, details: {})
+      error_summary = find_or_initialize_by(extraction_id: extraction_id)
+  
+      error_summary.extraction_name = extraction_name
+      error_summary.error_type = 'error'
+  
+      error_entry = {
+        message: message,
+        details: details,
+        timestamp: Time.current.iso8601,
+        worker_class: details[:worker_class],
+        job_id: details[:job_id],
+        harvest_job_id: details[:harvest_job_id],
+        pipeline_job_id: details[:pipeline_job_id],
+        harvest_report_id: details[:harvest_report_id],
+        stack_trace: details[:stack_trace],
+        context: details[:context] || {}
+      }
+  
+      error_summary.error_details = if error_summary.error_details.present?
+                                      error_summary.error_details + [error_entry]
+                                    else
+                                      [error_entry]
+                                    end
+  
+      error_summary.error_count = error_summary.error_details.size
+      error_summary.first_error_at ||= Time.current
+      error_summary.last_error_at = Time.current
+  
+      error_summary.save!
+      error_summary
+    end
+  
+    def self.log_stop_condition_hit(extraction_id:, extraction_name:, stop_condition_name:, stop_condition_content:, details: {})
+      error_summary = find_or_initialize_by(extraction_id: extraction_id)
+  
+      error_summary.extraction_name = extraction_name
+      error_summary.error_type = 'stop condition'
+  
+      # Determine if this is a system or user-defined stop condition
+      is_system_condition = details[:condition_type].present?
+  
+      error_entry = {
+        message: if is_system_condition
+                   "System stop condition '#{stop_condition_name}' was triggered"
+                 else
+                   "Stop condition '#{stop_condition_name}' was triggered"
+                 end,
+        details: details.merge(
+          stop_condition_name: stop_condition_name,
+          stop_condition_content: stop_condition_content,
+          is_system_condition: is_system_condition
+        ),
+        timestamp: Time.current.iso8601
+      }
+  
+      error_summary.error_details = if error_summary.error_details.present?
+                                      error_summary.error_details + [error_entry]
+                                    else
+                                      [error_entry]
+                                    end
+  
+      error_summary.error_count = error_summary.error_details.size
+      error_summary.first_error_at ||= Time.current
+      error_summary.last_error_at = Time.current
+  
+      error_summary.save!
+      error_summary
+    end
+  end
+  
--- a/app/sidekiq/delete_worker.rb
+++ b/app/sidekiq/delete_worker.rb
@@ -42,5 +42,35 @@ class DeleteWorker
     @harvest_report.update(delete_updated_time: Time.zone.now)
   rescue StandardError => e
     Rails.logger.info "DeleteWorker: Delete Excecution error: #{e}" if defined?(Sidekiq)
+    log_delete_worker_error(e, record, destination)
+  end
+
+  def log_delete_worker_error(exception, record, destination)
+    source_id = record.dig('transformed_record', 'source_id') ||
+                @harvest_report&.pipeline_job&.harvest_definitions&.first&.source_id ||
+                'unknown'
+    extraction_name = record.dig('transformed_record', 'job_id') ||
+                      @harvest_report&.pipeline_job&.harvest_definitions&.first&.name ||
+                      'unknown'
+
+    JobCompletionSummary.log_error(
+      extraction_id: source_id,
+      extraction_name: extraction_name,
+      message: "DeleteWorker error: #{exception.class} - #{exception.message}",
+      details: {
+        worker_class: self.class.name,
+        exception_class: exception.class.name,
+        exception_message: exception.message,
+        stack_trace: exception.backtrace&.first(20),
+        record: record,
+        destination_id: destination.id,
+        destination_name: destination.name,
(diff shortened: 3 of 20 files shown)


app/controllers/job_completion_summary_controller.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+class JobCompletionSummaryController < ApplicationController
+  before_action :set_job_completion_summary, only: [:show]
+
+  def index
+    @job_completion_summary = JobCompletionSummary.recent_errors.page(params[:page])
```
richardmatthewsdev, line 7: Can we keep the language consistent? So like JobCompletionSummary.recent_completions

app/controllers/job_completion_summary_controller.rb
```diff
@@ -0,0 +1,25 @@
...
+
+    if params[:error_type].present?
+      @job_completion_summary = @job_completion_summary.where(error_type: params[:error_type])
```
richardmatthewsdev, line 10: completion_type

app/models/job_completion_summary.rb
```diff
@@ -0,0 +1,104 @@
...
+  def self.group_by_error_type
+    group(:error_type).count
+  end
+
+  def self.total_errors
+    sum(:error_count)
+  end
+
+  def stop_condition?
+    error_type == 'stop condition'
```
richardmatthewsdev, line 24: completion_type

also probably good to store the type as a symbol without a space, so like `:stop_condition`

app/models/job_completion_summary.rb
```diff
@@ -0,0 +1,104 @@
...
+
+    error_summary.error_count = error_summary.error_details.size
+    error_summary.first_error_at ||= Time.current
+    error_summary.last_error_at = Time.current
+
+    error_summary.save!
+    error_summary
+  end
+
+  def self.log_stop_condition_hit(extraction_id:, extraction_name:, stop_condition_name:, stop_condition_content:, details: {})
```
richardmatthewsdev, line 68: These 2 methods are pretty large, I wonder if there is some way they can be improved?

app/sidekiq/delete_worker.rb
```diff
@@ -42,5 +42,34 @@ def delete(record, destination)
     @harvest_report.update(delete_updated_time: Time.zone.now)
   rescue StandardError => e
     Rails.logger.info "DeleteWorker: Delete Excecution error: #{e}" if defined?(Sidekiq)
+    log_delete_worker_error(e, record, destination)
+  end
+
+  def log_delete_worker_error(exception, record, destination)
```
richardmatthewsdev, line 48: Yeah it feels like the pattern for logging an error could be better

richardmatthewsdev: It could be good to if the code for logging of error wasn't part of the worker itself as well. So like the worker could just call 

something.log_delete_worker_error(pass some details)

and it would be wrapped up nicely

app/supplejack/extraction/execution.rb
```diff
@@ -48,7 +50,29 @@ def execution_cancelled?
     end
 
     def stop_condition_met?
```
richardmatthewsdev, line 52: I think instead of doing this you can put the logging of each condition within the method itself, then you don't need to hang onto the conditions or have the logic in here. 

So calling it as it is now

```
    def stop_condition_met?
      [set_number_reached?, extraction_failed?, duplicate_document_extracted?, custom_stop_conditions_met?].any?(true)
    end
```

then doing something like this for each condition

```
    def set_number_reached?
      return false unless @harvest_job.present? && @harvest_job.pipeline_job.set_number?
      
      log_stop_condition_hit('set_number_reached', 'Set number limit reached', { condition_type: 'set_number_reached' })

      @harvest_job.pipeline_job.pages == @extraction_definition.page
    end
```

app/supplejack/extraction/execution.rb
```diff
@@ -74,7 +98,30 @@ def custom_stop_conditions_met?
       stop_conditions = @extraction_definition.stop_conditions
       return false if stop_conditions.empty?
 
-      stop_conditions.map { |condition| condition.evaluate(@de.document.body) }.any?(true)
+      triggered_conditions = stop_conditions.select { |condition| condition.evaluate(@de.document.body) }
```
richardmatthewsdev, line 101: You might be able to put this in the evaluation of the stop condition rather than here, so you can evaluate like normal and if the block returns true you can log that it was hit

app/views/job_completion_summary/index.html.erb
```diff
@@ -0,0 +1,99 @@
...
+                  <th>Extraction Name</th>
+                  <th>Type</th>
+                  <th>Count</th>
+                  <th>First Occurrence</th>
+                  <th>Last Occurrence</th>
+                  <th>Actions</th>
+                </tr>
+              </thead>
+              <tbody>
+                <% @job_completion_summary.each do |error_summary| %>
```
richardmatthewsdev, line 37: this is the completion summary aye?

app/views/job_completion_summary/index.html.erb
```diff
@@ -0,0 +1,99 @@
...
+                                  class: 'btn btn-sm btn-outline-primary' %>
+                    </td>
+                  </tr>
+                <% end %>
+              </tbody>
+            </table>
+          </div>
+        <% else %>
+          <div class="text-center py-5">
+            <i class="bi bi-check-circle text-success" style="font-size: 3rem;"></i>
```
richardmatthewsdev, line 83: no inline styles please

app/controllers/job_completion_summary_controller.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+class JobCompletionSummaryController < ApplicationController
```
hapiben, line 3: **suggestion:** Plural `JobCompletionSummariesController`

richardmatthewsdev (changes requested):
Hey, it's looking good but I think it can be improved. The reek has also gone up quite a bit so it would be awesome if we can reduce that. In general I think if we can have some sort of class that handles the logging so the logic for logging isn't in all of the workers that would be nice. 

Let me know if you would like to chat about any of the comments :)

app/views/job_completion_summary/index.html.erb
hapiben, line 37: **suggestion:** Can we move this to a partial?

hapiben, line 37: **suggestion:** Expecting here to be plural `@job_completion_summaries`

app/views/job_completion_summary/index.html.erb
```diff
@@ -0,0 +1,99 @@
...
+                      <% end %>
+                    </td>
+                    <td>
+                      <span class="badge bg-warning text-dark">
+                        <%= error_summary.error_count %>
+                      </span>
+                    </td>
+                    <td>
+                      <small class="text-muted">
+                        <%= error_summary.first_error_at&.strftime('%Y-%m-%d %H:%M:%S') || 'N/A' %>
```
hapiben, line 63: **suggestion:** Can we use this? https://github.com/DigitalNZ/supplejack_harvester/blob/main/config/initializers/datetime_formats.rb

hapiben (changes requested):

spec/models/job_completion_summary_spec.rb
```diff
@@ -0,0 +1,262 @@
...
+      it 'filters by error type' do
+        expect(described_class.by_error_type('error')).to include(error_summary)
+        expect(described_class.by_error_type('error')).not_to include(stop_condition_summary)
+        expect(described_class.by_error_type('stop condition')).to include(stop_condition_summary)
+        expect(described_class.by_error_type('stop condition')).not_to include(error_summary)
+      end
+    end
+  end
+
+  describe 'class methods' do
```
hapiben, line 35: **suggestion:** Not sure the standards here, but having `.` on `describe` is enough to determine that it's a class method.

app/models/job_completion_summary.rb
```diff
@@ -0,0 +1,104 @@
...
+class JobCompletionSummary < ApplicationRecord
+  validates :extraction_id, presence: true, uniqueness: true
+  validates :extraction_name, presence: true
+  validates :error_type, presence: true
+  validates :error_details, presence: true
+  validates :error_count, presence: true, numericality: { greater_than_or_equal_to: 0 }
+
+  scope :by_error_type, ->(type) { where(error_type: type) }
+  scope :recent, -> { order(last_error_at: :desc) }
+  scope :recent_errors, -> { recent }
```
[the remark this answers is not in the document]
chigginsboost: good catch

app/models/job_completion_summary.rb
```diff
@@ -0,0 +1,143 @@
...
+  validates :job_type, presence: true
+
+  validates :source_id, uniqueness: { scope: %i[process_type job_type] }
+
+  validates :completion_entries, presence: true
+  validates :completion_count, presence: true, numericality: { greater_than_or_equal_to: 0 }
+
+  after_initialize :set_defaults, if: :new_record?
+
+  scope :recent_completions, -> { order(last_occurred_at: :desc) }
```
richardmatthewsdev, line 26: last completed at?

app/models/job_completion_summary.rb
```diff
@@ -0,0 +1,143 @@
...
+
+  def self.log_completion(params)
+    entry_params = if ['stop_condition', :stop_condition].include?(params[:completion_type])
```
richardmatthewsdev, line 29: Why do we have both the string and the symbol?

app/sidekiq/file_extraction_worker.rb
```diff
@@ -8,21 +8,41 @@ class FileExtractionWorker
 
   def perform(extraction_job_id)
     initialize_instance_variables(extraction_job_id)
+    reset_harvest_report_if_needed
```
richardmatthewsdev, line 11: you can probably drop the if_needed so it just calls it and it happens if it's required or not

app/views/job_completion_summaries/_summary_row.html.erb
```diff
@@ -0,0 +1,38 @@
...
+    <strong><%= error_summary.source_name %></strong>
+  </td>
+  <td>
+    <strong><%= error_summary.job_type %></strong>
+  </td>
+  <td>
+    <% if error_summary.stop_condition? %>
+      <% latest_entry = error_summary.completion_entries.last %>
+      <% if latest_entry&.dig('details', 'stop_condition_type') == 'system' %>
+        <span class="badge" style="background-color: #0d6efd;">System stop condition</span>
```
richardmatthewsdev, line 15: no inline styles :)

app/models/job_completion_summary.rb
chigginsboost: I cannot for the life of me remember why I did this but I will clean it up. Thanks richard

app/views/job_completion_summaries/_summary_row.html.erb
chigginsboost: My bad, I used Cursor to generate the UI and I think I missed a couple things when cleaning it up. Thanks!

app/models/job_completion_summary.rb
chigginsboost: doooooiiiiiii omg

app/controllers/job_completion_summaries_controller.rb
```diff
@@ -0,0 +1,28 @@
...
+
+    @job_completion_summaries = @job_completion_summaries.where(extraction_id: extraction_id)
+  end
+
+  def show; end
+
+  private
+
+  def set_job_completion_summary
+    @job_completion_summary = JobCompletionSummary.find(params[:id])
```
hapiben, line 26: **suggestion:** Looks like `show` is the only one who uses this. You can just put it there instead of having a callback method and a private method.

chigginsboost:
CODE QUALITY: Many of the reeks are occurring in files that were unrelated to the changes in this PR. We are going to revisit in another PR to clean up and refactor this code.

PR #157, opened 2025-11-28
[PROOF OF CONCEPT - TESTING ONLY, DO NOT MERGE] Pre-extraction to harvest nested sources 

chigginsboost:
Standalone, or 'independent' extractions makes harvesting content from nested pages possible. Instead of extracting records directly from a single URL, the system first traverses through link structures (tv show > season > episode) before fetching the actual content.

Screenshot 2025-12-10 at 4 34 41 PM

This story is a proof of concept. There are several follow-up stories to refine and improve this feature. 

**Overview** 

Currently, you add the selector for each page on the step template. Each step runs an extraction that only selects the links provided in the selector. Once it's finished, you should see output like this:

Screenshot 2025-12-15 at 6 33 31 PM

**How it works:**
- Configure an extraction definition with selectors for each page level (e.g., page 1: extract category links, page 2: extract item links)
- Create an automation with at least one independent extraction step, followed by a pipeline step
- The extraction job visits the source, extracts URLs at each depth, and saves them as JSON placeholders in folders.
- The final extraction job reads those saved URLs from the folders, fetches the content, and continues with the normal transformation/load flow 

**How to test**
UAT: https://harvester-uat-ch-pre-extraction.uat.digitalnz.org/pipelines
- Create a new pipeline, and then create an extraction. On the extraction definition, make sure that 'independent extraction' is set to Yes 
- run a sample extraction, create a transformation definition. make sure to add a field / code blocks to select part of the final data
- go to automation templates, and create a new step. make sure it is type 'independent extraction' and that it is pointing to the extraction definition that you created previously. Add a selector (xpath) to pick any links you want 
- repeat this for each level 
- when you've hit the last page that you would actually like to transform and load into the API, then make sure the automation step is type 'pipeline'. This will kick off the transformation and load. 
- make sure the transformation is pointing to the right extraction (not the test) 
- check the staging site or api pod logs to see the data is loaded correctly

 **Findings**
 
Findings and Next steps

There are two areas of improvement that we will address in the upcoming stories. First, we would like to create extraction independently, so they are not tied to a pipeline.

We should be able to create an extraction and extraction definitions from a different place (or places), possibly including the automation templates
- Extractions should have an optional pipeline ID.
- We should be able to preview extraction output from each place that we are able to create them
- We should streamline the UI so we don't have too much duplication and confusion on where or how to run an extraction, however, we might want a standalone page to view extractions

Second, the current POC doesn't utilize the existing functionality as well as we would like. There are several ways we could improve the process:
- The extraction jobs created by the automation steps should have their own extraction definitions
- The extractions should behave similar to the "preview" extraction by default. W
- The extraction definition should not have a bool flag that signifies that it's chained to another. It should have a previous_extraction_id so we can connect it to a previous job
- We should try to use the existing enrichment functionality and allow for a new parameter on an extraction (or perhaps utilize dynamic parameters) to pass through a block of code to select links. The parsing should happen from the code provided by the harvester, not something we need to account for.
- The current implementation is too focused on finding and parsing links. It would be better to have a more flexible solution that simply takes any data, and passes that data into a new extraction.
- All of the logic that currently lives on the automation templates should be removed, including the selector and any updates to the model.


app/controllers/extraction_definitions_controller.rb
```diff
@@ -15,10 +15,7 @@ def create
     @extraction_definition = ExtractionDefinition.new(extraction_definition_params)
 
     if @extraction_definition.save
-      @harvest_definition.update(extraction_definition_id: @extraction_definition.id)
-
-      2.times { Request.create(extraction_definition: @extraction_definition) }
-
+      setup_new_extraction_definition
```
richardmatthewsdev: since this has a flow on effect the convention would be to have a ! in the method name

app/models/extraction_job.rb
```diff
@@ -63,6 +64,47 @@ def documents
...
+
+    docs = documents
+    (1..docs.total_pages).filter_map { |page_number| extract_link_url(docs[page_number]) }
+  end
+
+  def extract_link_url(doc)
+    return nil unless doc
+
+    body = JSON.parse(doc.body)
+    body['url'] if body.is_a?(Hash) && body.keys == ['url'] && body['url'].present?
```
chigginsboost: *

app/sidekiq/harvest_worker.rb
```diff
@@ -17,37 +17,52 @@ def child_perform(harvest_job)
...
+      @pipeline_job.job_priority, (page * 5).seconds, @harvest_job.id, page
+    )
+    @harvest_report.increment_transformation_workers_queued!
+  end
 
   private
 
   def page_number_reached?(page)
     @pipeline_job.set_number? && page == @pipeline_job.pages
   end
```
chigginsboost: *

PR #166, opened 2026-02-26
Fix regression that put POST request body into the URL.

richardmatthewsdev:
When the POST request is appended to the URL the URL becomes much too long and gets rejected by Puma. This means that extractions that require sending large payloads do not work. 

This was originally fixed here

https://github.com/DigitalNZ/supplejack_harvester/pull/91

and was reintroduced here

https://github.com/DigitalNZ/supplejack_harvester/pull/131

### DIFF
--- a/app/supplejack/extraction/base_connection.rb
+++ b/app/supplejack/extraction/base_connection.rb
@@ -28,7 +28,8 @@ module Extraction
 
     private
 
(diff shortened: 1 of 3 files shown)


Mitsukiooka (approved):
LGTM!

motizuki (approved):

PR #169, opened 2026-03-30
Improve page load time for Supplejack

richardmatthewsdev:
Acceptance Criteria

The transformation definition page loads swiftly
The schedules page loads swiftly
Notes

At the moment it is taking roughly 20 seconds to load the transformation definition page. Candidates for this I think are the schema loading and the filesystem loading.

The schedules page overview has gotten very slow since we have added more schedules. It would be good to improve this as well, although this could be a seperate story.

What I have done:

Installed the Bullet gem https://github.com/flyerhzm/bullet and updated queries to have the appropriate includes to improve relational queries. 

For example, this has meant that loading the transformation definition page takes about 1.5 seconds instead of 34 seconds when you have a large Schema.


richardmatthewsdev:
There is a CVE that was picked up in bundle audit that requires upgrading devise from 4 -> 5. I will raise this with DNZ as I don't like doing major library updates as part of a bundle audit check.

danielaboost (approved):
This looks great @richardmatthewsdev, cool use of the bullet gem! Do we want to keep the config included for it? 

Would be a cool topic for a lightning talk 🌩️

richardmatthewsdev:
> This looks great @richardmatthewsdev, cool use of the bullet gem! Do we want to keep the config included for it?
> 
> Would be a cool topic for a lightning talk 🌩️

Thanks @danielaboost! Yes I want to keep it installed as it shows up in development mode and I think it will help catch some of these issues. It could be a cool lightning talk!

Mitsukiooka (approved):

PR #171, opened 2026-06-08
Stop silently dropping oversized extraction pages

hapiben:
NLNZcat OAI harvest was silently dropping page 4. No error, no records processed. Root cause: a `10 MB` guard in `RawRecordsExtractor` that returned `[]` for any page over that size without logging anything.

The limit was added in 2023 alongside the XML split feature, on the assumption split would keep pages under 10 MB. That assumption breaks when a page is large due to a single oversized record (16.7 MB MARC21 record in NLNZcat). Splitting by record count doesn't help.

**Changes:**
  - Raise threshold to 100 MB (NLNZcat page is 16.7 MB. Well within this)
  - Log a warning when threshold is exceeded instead of silently returning []
  - Update browser "too large to preview" message to clarify the page is still extracted and transformed
  - Remove stale UI copy advising users to "split" pages over 10 MB


richardmatthewsdev (approved):
Nice!

boost-tim (approved):

PR #176, opened 2026-07-06
Don't complete harvest reports while a split extraction is still pending

hapiben:
**Bug:** 
Split + fresh extraction → enrichment shows all 0s / "Completed" and never runs; re-running against the existing extraction works.

**Cause:** 
`ExtractionWorker#job_end` completes the harvest report right after download, while `SplitWorker` is still queued. All counters are 0 so the report cascades to "completed" and enrichments are enqueued before any records are loaded. The 0-record enrichment job then blocks re-enqueueing when the real load finishes.

**Fix:** 
Skip `extraction_completed!` when a split is pending . `SplitWorker` already marks it completed after splitting (same guard 1791145a added for text extraction).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -63,3 +63,6 @@ node_modules
 # from code editors
 /.vscode
 /app/frontend/jsconfig.json
+
+# LLMs
+.claude/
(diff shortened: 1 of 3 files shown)


richardmatthewsdev (approved):
Nice! Thanks @hapiben :)

motizuki (approved):

PR #177, opened 2026-07-07
Enrich every record when skip previously enriched is enabled

hapiben:
**Problem:**
- "Skip Previously Enriched" enrichments miss large blocks of records. A 100-record harvest only enriches ~60.
- Re-runs miss again: only about half of the remaining records get picked up each time.
- Small harvests (≤ 20 records) work fine.

**Root cause:**
- With skip enabled, the record query excludes records that already have the enrichment fragment (exclude_source_id).
- SjApiEnrichmentIterator pages through that query ascending, re-issuing it per page while the enrichment is creating those very fragments.
- Every enriched record drops out of the filtered results, shifting the records behind it into page positions already visited. Those records are never fetched.
- With the API's fixed page size of 20: pages 1–20 enrich, "page 2" now returns originals 41–60 (21–40 slid into page 1's spot), etc. → exactly 60/100.

**Fix:**
- Process pages last-to-first. Enriched records always sit behind the pages still to be fetched, so nothing shifts under the iterator.
- Correct regardless of when fragments land. Immune to the timing race.
- Sample jobs and cancellation behaviour unchanged.

### DIFF
--- a/app/supplejack/extraction/sj_api_enrichment_iterator.rb
+++ b/app/supplejack/extraction/sj_api_enrichment_iterator.rb
@@ -12,15 +12,20 @@ module Extraction
       @max_page = find_max_page(@first_api_document) if @first_api_document.body.present?
     end
 
+    # Enriching a record removes it from the 'skip previously enriched'
+    # filtered results, shifting every record behind it forward a page.
+    # Pages are processed from last to first so that the shrinking result
+    # set never shifts records into a page that has already been processed.
     def each
-      yield(@first_api_document, @page)
-      @page += 1
+      if @max_page.present?
+        @max_page.downto(@page + 1).each do |page|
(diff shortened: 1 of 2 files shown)


richardmatthewsdev:
Hey bro, I appreciate that this works and that is an odd problem that I didn't think about at the time. I feel like a larger but better fix would be to improve the way the enrichment pagination in the API works so instead of doing a query each time you could have a cursor based pagination approach so you aren't at risk of the results changing while you are paging through the enrichment. That would also help other scenarios where new records were added to the collection while an enrichment is running and get missed.

richardmatthewsdev (approved):

hapiben:
Thanks @richardmatthewsdev. I'll have a story to address the cursor-based pagination.

PR #178, opened 2026-07-16
Pre-processing pipeline

hapiben:
This adds a new kind of block called a **pre-processing block**. These blocks run first, before the harvest. You can add many of them, and they run one after another in order. Each block passes its result to the next block.

**Changes/What's new**
- A new **"preprocess"** block type, with an order number so blocks run in the right sequence.
- Blocks run one at a time, in order:
  - The first block gets its own data from the source.
  - The next blocks use the output of the block before them, instead of pulling from the source again.
- Each block **saves its output to disk** and passes it to the next block. It still works even if some pages are empty.
- New buttons on the pipeline page to **add and manage** pre-processing blocks.
- Enrichment jobs do **not** start when a pre-processing block finishes.

Screenshot 2026-07-21 at 10 50 47 AM
Screenshot 2026-07-21 at 10 51 10 AM


paul-mesnilgrente (commented):
It's pretty big but it looks goog to me.

paul-mesnilgrente (approved):

richardmatthewsdev (approved):
Hey bro, I think this is looking good. My only concern is how it overlaps with the automations feature. 

The intention with automations was to be able to run things in a specific order and pass details between jobs which is very similar to what the pre process does. 

I feel like at some point it would be worth working with Nick to see if ideas from automations and pipelines can be combined together somehow so we have a single way of running the steps. 

The starting point for the harvester was just having extractions, transformations, and loads, and you run them in any way you wanted, and then we added the pipeline concept around it so that it would make more sense but I feel like both pipelines and automations aren't quite right yet and there is a risk that they just sort of merge in a funny way. Not something for this story but maybe something to think about in future :).

motizuki (approved):

PR #180, opened 2026-07-21
Draft: Pre-processing output viewer

hapiben:
Adds a read-only viewer so you can inspect the data a pre-processing block feeds into the harvest, to confirm the pipeline is configured correctly.

- A "View pre-processed data" button on each pre-processing block card.
- An index page listing the pipeline runs that produced output for that block (status, date), and a show page that pages through the output records using the same JSON viewer as extracted data.
- View-only by design

Screenshot 2026-07-28 at 11 05 30 AM


danielaboost (approved):
Nice one, this is a cool feature @hapiben!

motizuki (approved):

PR #182, opened 2026-07-23
Upgrade Harvester Rails

paul-mesnilgrente:
**Acceptance Criteria**
- Rails is updated with the next supported version on SJ Harvester

Support for Rails 7.2.3.1 ends on 2026-08-09


public/404.html
```diff
@@ -1,67 +1,114 @@
...
+        -webkit-text-size-adjust: 100%;
+      }
+
+      a {
+        color: inherit;
+        font-weight: 700;
+        text-decoration: underline;
+        text-underline-offset: 0.0925em;
+      }
+
```
chigginsboost, line 50: nit: I think all these pages are generated and probably not something we want to own or manage, but if not it might be nice to move the css an svgs into separate files. I'm not too fussed about this, though :)

chigginsboost (approved):

public/404.html
paul-mesnilgrente: I would agree, but we currently don't overwrite them. So I think keeping the Rails defaults here is sensible. We can remove these files when we start overwriting error pages.

PR #237, opened 2026-08-03
Part 3: "Run" modal with pre-processed data

paul-mesnilgrente:
Before:
image

After:
image


chigginsboost:
I think it looks good Paul!

